PR #15, opened 2024-03-21
releng: Upgrade target to dependencies of 2024-03 (Eclipse 4.31) release

PatrickTasse:
Trace Compass 9.3.0


bhufmann (approved):
I did a build on my computer and test the Incubator TC RCP and the Trace Server. Both work fine. Thanks for the upgrade.

Issue #12, opened 2024-03-19
Remove incubator flamechart views from trace server


bhufmann:
With the migration of the incubator flame chart and related views as well as their implementation to mainline Trace Compass https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/9 the trace server will now show 3 different flame chart views (classic, incubator and new) in the available views. As a first step the incubator flamechart plug-in should be removed from the trace server.

Removing the flamechart plug-in will require other incubator plug-ins to remove their dependencies to the incubator flamechart plug-in and use the mainline implementation instead.

### Plug-ins to remove from traceserver.product
- org.eclipse.tracecompass.incubator.callstack.core
- org.eclipse.tracecompass.incubator.analysis.core

### Plug-ins to update
- org.eclipse.tracecompass.incubator.rocm.core
-  org.eclipse.tracecompass.incubator.traceevent.core
- org.eclipse.tracecompass.incubator.traceevent.core.tests

Note: The source code migrated to the mainline Trace Compass should be removed from the incubator repository.

arfio:
I agree that this should be done after the following pull request is merged: 
https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/23
Aside from that: 
- I have updated the ROCm core plugin, I need to push the changes and make a PR for it.
- The other plugins should follow.

bhufmann:
I noticed that `org.eclipse.tracecompass.incubator.callstack.core.sampled.callgraph` package hasn't been migrated to mainline. This needs to be done before removing the incubator plug-in.

arfio:
The `org.eclipse.tracecompass.incubator.callstack.core.sampled.callgraph` package was added to this PR: https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/23

PR #18, opened 2024-04-04
Remove references to the incubator callstack

arfio:
This PR is intended to resolve the issue https://github.com/eclipse-tracecompass-incubator/org.eclipse.tracecompass.incubator/issues/12

Some classes were not moved to profiling.core so this patch is dependent on https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/51

This fixes the following projects:
- org.eclipse.tracecompass.incubator.traceevent.core
- org.eclipse.tracecompass.incubator.traceevent.core.tests
- org.eclipse.tracecompass.incubator.inandout.core
- org.eclipse.tracecompass.incubator.kernel.core
- org.eclipse.tracecompass.incubator.tracecompass.core
- org.eclipse.tracecompass.incubator.uftrace.core
- org.eclipse.tracecompass.incubator.virtual.machine.analysis.core
- org.eclipse.tracecompass.incubator.virtual.machine.analysis.core.tests
- org.eclipse.tracecompass.incubator.kernel.ui
- org.eclipse.tracecompass.incubator.tracecompass.core
- org.eclipse.tracecompass.incubator.trace.server.product
- org.eclipse.tracecompass.incubator.perf.profiling.core
- org.eclipse.tracecompass.incubator.perf.profiling.core.tests
- org.eclipse.tracecompass.incubator.otf2.core
- org.eclipse.tracecompass.incubator.scripting.core
- org.eclipse.tracecompass.incubator.scripting.ui

[Fixes] remove incubator callstack dependencies


bhufmann (changes requested):
I noticed that the ROCM plugins still depends on the incubator callstack. Additionally, the trace server also includes the incubator.core plug-in. Both should be addressed as well. 

Could you please update that or let me know if have a different plan for those? 

Thanks
Bernd

bhufmann:
Trace Compass baseline build is available and this PR doesn't have any compilation errors anymore. However, there are some test errors which need to be addressed as well:

https://github.com/eclipse-tracecompass-incubator/org.eclipse.tracecompass.incubator/actions/runs/8559375092/job/23678280523?pr=18#step:5:14103

arfio:
The ROCm plugin will be updated separately, the incubator callstack and the ROCm plugin were removed from the trace server. The incubator analysis was kept because the trace event plugin is dependent on the aspects.
The tests errors were corrected.

bhufmann (changes requested):
I'm ok to handle ROCm in a separated PR.

I noticed that the feature.xml files still have the dependencies to the incubator callstack plug-ins listed. I think they should be removed as will as part of the PR.

While checking the feature.xml files. I noticed that the XML defined call stack have not been migrate to mainline (package: . `org.eclipse.tracecompass.incubator.internal.callstack.core.xml.callstack`). The plug-in `org.eclipse.tracecompass.incubator.tracecompass.core` defines a build-in xml analysis (`GroupedTasks.xml`)  and `org.eclipse.tracecompass.incubator.traceevent.core` defines another one (`GroupedTasksTraceEvent.xml`). These 2 corresponding features.xml cannot have the incubator callstack dependencies removed just yet. 

So, I suggest to handle the xml analysis in a separate PR because it needs to be migrated first to mainline Trace Compass. However, please update this PR to update all relevant feature.xml files. WDYT?

arfio:
Good catch, I agree with you. Do you think the xml analysis should be added to org.eclipse.tracecompass.analysis.profiling.core or org.eclipse.tracecompass.tmf.analysis.xml.core ? I personally think xml would be better.

bhufmann:
> Good catch, I agree with you. Do you think the xml analysis should be added to org.eclipse.tracecompass.analysis.profiling.core or org.eclipse.tracecompass.tmf.analysis.xml.core ? I personally think xml would be better.

I think it should be part of `org.eclipse.tracecompass.tmf.analysis.xml.core` with the rest of the xml analysis code.

bhufmann (approved):
Looks good to me. Thanks!

PR #20, opened 2024-04-10
Update ROCm to be more generic

arfio:
This change will update the ROCm plugin to be more generic.
It adds a GPU plugin that generalizes the ROCm analysis and a layout definition to ROCm.

[Added] GPU plugin
[Added] RocmApiEventLayout and other classes to work with the GPU plugin
[Fixed] Dockerfile alpine version


MatthewKhouzam (approved):
Looks good

PR #21, opened 2024-04-11
xml-analysis: remove reference to nashorn

arfio:
This fixes warnings about the nashorn engine missing.


bhufmann (approved):
Thanks for this. Nashorn scripting engine doesn't exist anymore in recent JVM versions. Rhino from Mozilla is now used per default.

PR #22, opened 2024-04-16
callstack: remove xml analysis

arfio:
This PR removes the xml analysis from the incubator so that it does not create any conflict with the one from https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/61


bhufmann (approved):
Thanks for this update. Make sense to remove it in conjunction with addition of this feature to mainline Trace Compass.

arfio:
The addition of this feature to mainline Trace Compass was breaking the build of the incubator so this change was necessary.

PR #24, opened 2024-04-21
ros2: support pub-sub message tracking using rmw layer only

christophebedard:
The current implementation of the ROS 2 plugin relies on events from the underlying DDS middleware in order to be able to track a message from publication to subscription: it needs to get a timestamp value from DDS when a message is published. This timestamp value is then matched to the same value when the message is received on the other end.

While the ROS 2 instrumentation is available by default on Linux, instrumentation for the DDS implementation(s) is not available by default; the DDS implementation needs to be built from source. However, some recent changes to the ROS 2 instrumentation mean that DDS instrumentation is not needed anymore in order to track messages from publication to subscription. We just rely on events from the ROS 2 `rmw` layer (which is right above DDS in the ROS 2 architecture), specifically the `ros2:rmw_publish` events, to get the same timestamp. For more information about these changes, see: https://github.com/ros2/ros2_tracing/pull/74.

Therefore, note the version of the instrumentation (aka the ROS 2 `tracetools` package) used to generate the trace. If the version is greater than or equal to the version that contains the new changes, rely on `rmw` for the timestamp value. Otherwise, if an older version is used, rely on DDS for the timestamp value.


MatthewKhouzam (approved):

christophebedard:
Thanks for the approval. I'm wary about merging this after a few months and just before leaving on vacation (away from any computer) for 1 week, so I'll merge this next Friday.

christophebedard:
I rebased this. Looks like the approval is still valid, so I'll go ahead and merge this.

PR #29, opened 2024-05-23
Fix stable 9 3 update site

bhufmann:
- Upgrade target of stable-9.3 branch to e4.31
- Add missing dependencies on incubator update site to allow installing incubator features from update-site

 

common/org.eclipse.tracecompass.incubator.releng-site/category.xml
```diff
@@ -89,6 +89,9 @@
    <bundle id="org.eclipse.wst.jsdt.manipulation"/>
    <bundle id="org.eclipse.nebula.widgets.opal.commons"/>
    <bundle id="org.eclipse.nebula.widgets.opal.duallist"/>
+   <bundle id="jakarta.activation-api" version="1.2.2"/>
+   <bundle id="jakarta.xml.bind-api" version="2.3.3"/>
+   <bundle id="javax.inject"/>
```
PatrickTasse, line 94: I worry if this should be jakarta.inject.jakarta.inject-api in order to get the same version than in orbit-aggregation?

common/org.eclipse.tracecompass.incubator.target/tracecompass-incubator-master.target
```diff
@@ -131,7 +131,7 @@
 <unit id="org.eclipse.tracecompass.lttng2.control.feature.group" version="0.0.0"/>
 <unit id="org.eclipse.tracecompass.jsontrace.feature.group" version="0.0.0"/>
 <unit id="org.eclipse.tracecompass.tmf.cli.feature.group" version="0.0.0"/>
-<repository location="https://download.eclipse.org/tracecompass/master/repository/"/>
+<repository location="https://download.eclipse.org/tracecompass/releases/9.3.0/repository/"/>
```
PatrickTasse, line 134: Does this mean all stable branches from incubator must diverge from master branch from now on? They haven't usually.

bhufmann: all stable branches don't build because they reference tc master. We would have to change the targets in ALL stable branches. But I think it's overkill. We can do it on need base.

common/org.eclipse.tracecompass.incubator.releng-site/category.xml
bhufmann: I'll try that.

bhufmann: I'm not able to make it work with this one. Maybe it's ok to use as-is. Soon we have the 10.0 version.

bhufmann: For some reason plug-in `org.eclipse.ease.modules.platform` needs exact version `1.0.0` of `javax.inject`. Here that's what we get from the MANIFEST.mf:
```
`javax.inject;bundle-version="1.0.0"`
```

PR #48, opened 2024-07-03
releng: update tracecompass-test-traces repository

arfio:


bhufmann (changes requested):
Please update the sequence number at the beginning of the target file. Also update the tracecompass-incubator-e4.20.target file the same way.

bhufmann (approved):
Looks good to me. Thanks!

PR #50, opened 2024-07-10
Table column data type

kavehshahedi:
Added data type to the event and segment aspects (Segment already had getType, but the aspects didn't). The aim was to synchronize the trace-server functionalities w.r.t. tables in order to indicate the type of columns.

### DIFF
--- a/trace-server/README.md
+++ b/trace-server/README.md
@@ -13,7 +13,7 @@
 
 ## Compiling manually
 
-To build the project manually follow the instructions in [BUILDING](#compiling-manually).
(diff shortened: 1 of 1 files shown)


bhufmann:
@kavehshahedi Since https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/123 is merged into Trace Compass mainline, we can review this PR. However, you need to rebase it to the latest in master branch. Please note that an additional plug-ins were added that provides an event aspect (o.e.tc.incubator.gpu.core, o.e.tc.incubator.jifa.core, o.e.tc.incubator.dpdk.core). Please update them as well if needed. Thanks.

PR #52, opened 2024-07-12
server: Fix potential trace UUID mismatch when creating experiment model

### LINKED ISSUE #51: Potential trace UUID mismatch in experiments
When creating the experiment TSP model may have trace UUID mismatch if the children traces have also children. Basically, it's not guaranteed  that the order of traces returned by `TmfTraceManager.getTraceSet(experiment)` is the same order than the order returned by the iterator `iter` containing the UUIDs of the traces, and hence the UUID of a trace could be mismatched. The following code needs to update to make sure that the traces have the correct UUIDs assigned. 

```java
public static Experiment from(TmfExperiment experiment, UUID expUUID) {
        Iterator<UUID> iter = ExperimentManagerService.getTraceUUIDs(expUUID).iterator();
        // Get all the leaf traces from experiment
        @NonNull List<@NonNull ITmfTrace> children = new ArrayList<>(TmfTraceManager.getTraceSet(experiment));
        Set<Trace> traces = Sets.newLinkedHashSet(Lists.transform(children,
                t -> Trace.from(t, iter.next())));
        return new Experiment(experiment.getName(),
                expUUID,
                experiment.getNbEvents(),
                experiment.getStartTime().toNanos(),
                experiment.getEndTime().toNanos(),
                experiment.isIndexing() ? "RUNNING" : "COMPLETED", //$NON-NLS-1$ //$NON-NLS-2$
                traces);
    }
```

This line causes the potential mismatch:

https://github.com/eclipse-tracecompass-incubator/org.eclipse.tracecompass.incubator/blob/4b14d71637df2cedecd48d4287851315418a2d31/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/Experiment.java#L94

bhufmann:
Don't rely on the order of traces array in experiment and trace UUIDs stored in ExperimentManagerService to create TSP Experiment model. Instead use the trace resources to create the TSP Trace model return in the TSP Experiment Model.

Fixes #51

[Fixed] potential trace UUID mismatch when creating experiment model

### DIFF
--- a/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/Experiment.java
+++ b/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/Experiment.java
@@ -12,23 +12,17 @@
 package org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core.services;
 
 import java.io.Serializable;
-import java.util.ArrayList;
-import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.UUID;
 
 import org.eclipse.core.resources.IResource;
-import org.eclipse.jdt.annotation.NonNull;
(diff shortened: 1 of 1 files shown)


MatthewKhouzam (approved):
Wow, nice reduction!

PR #55, opened 2024-07-19
execution-comparison: add to RCP

vladarama:
This PR adds the execution-comparison feature to the incubator's RCP.

Signed-off-by: Vlad Arama <vlad.arama@ericsson.com>


rcp/org.eclipse.tracecompass.incubator.rcp/feature.xml
MatthewKhouzam, line 1: I am curious, why is junit needed for an RCP?

vladarama: The build was failing due to a missing requirement. 
Not sure if my change is correct, but it fixed the failing build.

bhufmann (approved):
LGTM. Thanks!

MatthewKhouzam (approved):
That's more like it!

PR #57, opened 2024-08-21
PoC: Endpoint to download JSON schema as application:octet-stream

bhufmann:
PoC: Endpoint to download JSON schema as application:octet-stream

For demonstration purposes


bhufmann:
Closing since just a PoC

PR #63, opened 2024-08-28
Remove UI dependencies in various core plug-ins

bhufmann:
Core plug-ins should not have UI dependencies. 


MatthewKhouzam (commented):
This was to have icons. maybe we should document it.

bhufmann:
> This was to have icons. maybe we should document it.

Where are the icons referenced?

MatthewKhouzam:
The UI was used to give the icons for the trace types. Removing the dependencies actually has no negative impact in eclipse, it still works. This will reduce the burden on the trace server side. Thank you!

MatthewKhouzam (approved):

PR #65, opened 2024-08-28
uftrace: fix symbol provider to not fix false negatives

MatthewKhouzam:
The symbol provider now returns null when it cannot resolve rather than returning a string address. This allows chaining of symbol resolvers.

It is also made into a static class so that it is easier to debug.

Change-Id: Ifdfb7877f623c5e96feae0138308544ad4c0ab81


arfio (approved):
I tested it, it works well, the return null is already documented in the interface. LGTM !

MatthewKhouzam:
Thanks!

PR #69, opened 2024-09-05
trace-event: improve error message

MatthewKhouzam:
Give a full message instead of "Not a trace event trace" Also replace method reference instead of lambda.

Change-Id: I35b6d68b9ba66422a34a955ed9036cc01ff48acd


christophebedard (approved):
A lot more descriptive!

bhufmann (approved):

PR #73, opened 2024-09-12
rocm: validate only ust traces containing rocm events

arfio:
The precedent validation would identify every lttng-ust traces as a rocm exatracer trace. This patch corrects that by looking if the trace has any events from a hip or hsa provider.


MatthewKhouzam (approved):

bhufmann (approved):
Looks good to me. Thanks!

bhufmann:
I'm going to merge it so that it's included with the trace server.

PR #74, opened 2024-09-16
[server][readme] fix reference to general build instructions

marcdumais-work:
Currently the reference points to itself - it seems to make more sense to point to the root BUILDING.md file.


bhufmann (approved):
Looks good to me. Thanks!

PatrickTasse (approved):

PR #79, opened 2024-09-20
[server] Add trace properties to the Trace instance

siwei-zhang-work:
[server] Add trace properties to the Trace instance

The purpose of this commit is to get the trace properties using tsp. The "properties" attribute will be in the trace instance, and it will be filled if the opened trace implements ITmfPropertiesProviders.

Example:
{
  "name": "string",
  "path": "string",
  "nbEvents": 0,
  "start": 0,
  "indexingStatus": "RUNNING",
  "end": 0,
  "properties":
  {
	"hostname": "qemu1",
        "clock_offset": "1450192743562703624"
  },
  "UUID": "f50af7e0-0dd5-4361-ab96-2e04f7bc7e30"
}

[Added] "properties" field in the returned Trace instance
[Added] a separate trace instrance will be created when "/traces" is queried
[Changed] /experiment endpoint will use Trace.from(ITmfTrace) to get more info of the trace

### DIFF
--- a/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/model/Trace.java
+++ b/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/model/Trace.java
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (c) 2021 Ericsson
+ * Copyright (c) 2021, 2024 Ericsson
  *
  * All rights reserved. This program and the accompanying materials are
  * made available under the terms of the Eclipse Public License 2.0 which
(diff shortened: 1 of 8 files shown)


siwei-zhang-work:
Creating a separate trace instance doesn't feel like a clean solution, but I haven't found a better workaround. The issue is that when the experiment instance is disposed, it also disposes of the trace instance, but the TSP requires the trace instance to persist. Do you have any alternative suggestions for handling this?

MatthewKhouzam (approved):
I like it.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/model/Trace.java
bhufmann, line 1: This class should also be updated because it's used with swagger to create the TSP API doc. Could you please update it?

Also, after that could you please also generate a new API and upload it as PR to the trace-server-protocol repository?
https://github.com/eclipse-cdt-cloud/trace-server-protocol?tab=readme-ov-file#generate-the-api

bhufmann (changes requested):
Thanks for this contribution. @siwei-zhang-work small comment.

bhufmann (approved):
Looks good to me. Thanks for the contribution. I'll merge it after a successful build.

bhufmann:
I'll cherry-pick it to master.

PR #84, opened 2024-09-27
Return Table Column's Data Type (e.g., STRING, NUMBER, TIMESTAMP, etc.)

kavehshahedi:
When requesting the table columns, the server will now return the type of each column. The types are derived from `DataType` enum, such as `STRING`, `NUMBER`, `TIMESTAMP`, `DURATION`, etc.

This feature supports both `IEventAspect` and `ISegmentAspect` data providers.


MatthewKhouzam (approved):
I am good with this.

PR #80, opened 2024-09-23
server: support for customizable data providers (outputs) via data provider factories

bhufmann:
It uses the new `IDataProviderFactory.getAdapter()` interface to get `ITmfDataProviderConfigurator` implementation to create data providers. It passes an `ITmfConfiguration` instance with default TSP parameters (name, description, typeId) and the custom parameters in a JSON converted `Map<String, Object>` on to the configurator which the implementer of `ITmfDataProviderConfigurator` needs to apply.

It depends on 
https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/154
https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/167

~~This PR includes changes of PR #71 and its dependent pull requests, and will be rebased once dependent PRs are merged.~~

PR #71, opened 2024-09-09
server: Skeleton code for configuration service per data provider

bhufmann:
server: Skeleton code for configuration service per data provider

[Added] Skeleton code for configuration service per data provider

Based on commits of PR #70. This one will be rebased once it's PR #70 merged.

Signed-off-by: Bernd Hufmann <bernd.hufmann@ericsson.com>

PR #70, opened 2024-09-06
server: Update query parameters for configurations

bhufmann:
~~PR based on PR #66, which commits are included.~~

An additional commit is added:

- server: Update query parameters for configurations

This commit also includes improved swagger updates of related topics.

Signed-off-by: Bernd Hufmann <bernd.hufmann@ericsson.com>

PR #66, opened 2024-08-30
Support for json-schema in configuration source type

bhufmann:
Support for [json-schema](https://json-schema.org/) in configuration source type:

- server: Serialize TmfConfigurationSourceType schema as string
- server: add test for TmfConfigurationSourceType with schema as string
- swagger: Document schema in TmfConfigurationSourceType

Requires https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/141

See also
- https://github.com/eclipse-cdt-cloud/theia-trace-extension/pull/1108
- https://github.com/eclipse-cdt-cloud/trace-server-protocol/pull/103

Signed-off-by: Bernd Hufmann <bernd.hufmann@ericsson.com>

PR #88, opened 2024-10-01
Custom analysis API and InAndOut Implementation

bhufmann:
**Update (Oct 28, 0204):**
All the PRs below as well as relevant Trace Compass mainline are merged. This PR only has the changes of the InAndOut implementation left.

-------

This PR includes the TSP updates and concrete implementation for customizable analysis. The InAndOutAnalysis has been updated for that. This PR combines multiple commits, where some of them are available as separate PRs. It requires the commits of the following Trace Comapss mainline PR:

https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/164

This PR combines the following commits/PRs:

- #66
  - swagger: Document schema in TmfConfigurationSourceType
  - server: Serialize TmfConfigurationSourceType schema json object
- #70 
- #71  
- #80
- InAndOutAnalysis with data provider factory and json string

Additional related changes:
- #97 
- #98
- #100
- #103
- #110

### DIFF
--- a/analyses/org.eclipse.tracecompass.incubator.inandout.core/schema/in-and-out-analysis.json
+++ b/analyses/org.eclipse.tracecompass.incubator.inandout.core/schema/in-and-out-analysis.json
@@ -1,0 +1,59 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://org.eclipse.tracecompass/in-and-out-analysis.json",
+    "title": "InAndOut Analysis",
+    "description": "Custom Execution Analysis schema",
+    "type": "object",
+    "properties": {
+        "name" : {
+            "type": "string",
+            "description": "The name of this InAndOut configuration"
+        },
+        "description" : {
+            "type": "string",
+            "description": "The descrition of this InAndOut configuration"
+        },
+        "specifiers": {
+            "description": "array specifiers",
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "label": {
+                        "type": "string",
+                        "descritpion": "The label of the identifier"
+                    },
+                    "inRegex": {
+                      "type": "string",
+                      "descritpion": "The IN regular expression"
+                    },
+                    "outRegex": {
+                     "type": "string",
+                     "descritpion": "The OUT regular expression"
+                    },
+                    "contextInRegex": {
+                     "type": "string",
+                     "descritpion": "The context IN regular expression"
+                    },
+                    "contextOutRegex": {
+                     "type": "string",
+                     "descritpion": "The context OUT regular expression"
+                    },
+                    "classifier": {
+                     "type": "string",
+                     "descritpion": "The classifier"
+                    },
+                    "required": [
+                        "label",
+                        "inRegex",
+                        "outRegex",
+                        "contextInRegex",
+                        "contextOutRegex",
+                        "classifier"
+                    ]
+               }
+            }
+        }
+    },
+    "required": ["name", "specifiers"]
+}
--- a/analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutAnalysisModule.java
+++ b/analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutAnalysisModule.java
@@ -23,9 +23,18 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Path;
 import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.tracecompass.analysis.profiling.core.instrumented.InstrumentedCallStackAnalysis;
 import org.eclipse.tracecompass.incubator.internal.inandout.core.Activator;
+import org.eclipse.tracecompass.tmf.core.analysis.IAnalysisModule;
+import org.eclipse.tracecompass.tmf.core.config.ITmfConfiguration;
+import org.eclipse.tracecompass.tmf.core.config.TmfConfiguration;
+import org.eclipse.tracecompass.tmf.core.exceptions.TmfAnalysisException;
+import org.eclipse.tracecompass.tmf.core.exceptions.TmfConfigurationException;
+import org.eclipse.tracecompass.tmf.core.exceptions.TmfTraceException;
 import org.eclipse.tracecompass.tmf.core.statesystem.ITmfStateProvider;
 import org.eclipse.tracecompass.tmf.core.trace.ITmfTrace;
 import org.eclipse.tracecompass.tmf.core.trace.TmfTraceManager;
@@ -50,6 +59,42 @@ public class InAndOutAnalysisModule extends InstrumentedCallStackAnalysis {
      */
     public static final String JSON = ".config.json"; //$NON-NLS-1$
 
+    private @Nullable ITmfConfiguration fConfiguration;
+    private @Nullable List<@NonNull SegmentSpecifier> fSpecifiers;
+
+    @Override
+    public @NonNull String getName() {
+        ITmfConfiguration config = fConfiguration;
+        if (config != null) {
+            return config.getName();
+        }
+        return super.getName();
+    }
+
+    @Override
+    public @NonNull String getHelpText() {
+        ITmfConfiguration config = fConfiguration;
+        if (config != null) {
+            return config.getDescription();
+        }
+        return super.getHelpText();
+    }
+
+    @Override
+    public void setConfiguration(ITmfConfiguration configuration) throws TmfConfigurationException {
+        if (configuration == null) {
+            throw new TmfConfigurationException("Can't set a null configuration"); //$NON-NLS-1$
+        }
+        SegmentSpecifierConfiguration config = SegmentSpecifierConfiguration.fromJsonMap(configuration.getParameters());
+        fConfiguration = configuration;
+        fSpecifiers = config.getSpecifiers();
+    }
+
+    @Override
+    public @Nullable ITmfConfiguration getConfiguration() {
+        return fConfiguration;
+    }
+
     /**
      * Reference (default) specifier configured.
      */
@@ -57,16 +102,26 @@ public class InAndOutAnalysisModule extends InstrumentedCallStackAnalysis {
 
     @Override
     public @NonNull String getId() {
-        return ID;
+        ITmfConfiguration config = fConfiguration;
+        if (config == null || fSpecifiers == null) {
+            return ID;
+        }
+        return ID + config.getId();
     }
 
     @Override
     public boolean canExecute(@NonNull ITmfTrace trace) {
-        File config = getConfig(trace);
-        return config.exists() && super.canExecute(trace);
+        ITmfConfiguration configuration = fConfiguration;
+        if (configuration == null) {
+            // Eclipse Trace Compass
+            File config = getConfig(trace);
+            return config.exists() && super.canExecute(trace);
+        }
+        return true;
     }
 
     private static File getConfig(@NonNull ITmfTrace trace) {
+        // Eclipse Trace Compass
         String folder = TmfTraceManager.getSupplementaryFileDir(trace);
         File config = new File(folder + File.separator + ID + JSON);
         if (!config.exists()) {
@@ -80,9 +135,13 @@ public class InAndOutAnalysisModule extends InstrumentedCallStackAnalysis {
     @Override
     protected @NonNull ITmfStateProvider createStateProvider() {
         ITmfTrace trace = Objects.requireNonNull(getTrace(), "Trace should not be null at this point"); //$NON-NLS-1$
-        File configFile = getConfig(trace);
-        List<@NonNull SegmentSpecifier> list = read(configFile);
-        return new InAndOutAnalysisStateProvider(trace, list);
+        List<@NonNull SegmentSpecifier> specifiers = fSpecifiers;
+        if (specifiers == null) {
+            // Eclipse Trace Compass
+            File configFile = getConfig(trace);
+            specifiers = read(configFile);
+        }
+        return new InAndOutAnalysisStateProvider(trace, specifiers);
     }
 
     /**
@@ -116,10 +175,118 @@ public class InAndOutAnalysisModule extends InstrumentedCallStackAnalysis {
      *            the specifiers
      */
     public static void write(File file, List<@NonNull SegmentSpecifier> specifiers) {
+        // Eclipse Trace Compass
         try (Writer writer = new FileWriter(file)) {
             writer.append(new Gson().toJson(specifiers));
         } catch (IOException e) {
             Activator.getInstance().logError(e.getMessage(), e);
         }
     }
+
+    /**
+     * Create the InAndOutAnalysisModule for a given configuration and trace
+     *
+     * @param config
+     *            the input {@link ITmfConfiguration}
+     * @param trace
+     *            the trace to apply it to
+     * @param writeConfig
+     *            write the config (do only once)
+     * @return InAndOutAnalysisModule
+     * @throws TmfConfigurationException
+     *             if an error occurs
+     */
+    public static InAndOutAnalysisModule create(@NonNull ITmfConfiguration config, @NonNull ITmfTrace trace, boolean writeConfig) throws TmfConfigurationException {
+        /*
+         * Apply configuration to each trace (no need to check trace type here)
+         */
+        InAndOutAnalysisModule module = new InAndOutAnalysisModule();
+        module.setConfiguration(config);
+        if (writeConfig) {
+            IPath traceConfigPath = getTraceRootFolder(trace, config.getSourceTypeId());
+            TmfConfiguration.writeConfiguration(config, traceConfigPath);
+        }
+        try {
+            if (module.setTrace(trace)) {
+                IAnalysisModule oldModule = trace.addAnalysisModule(module);
+                if (oldModule != null) {
+                    oldModule.dispose();
+                    oldModule.clearPersistentData();
+                }
+            } else {
+                module.dispose();
+                throw new TmfConfigurationException("InAndOut analysis module can't be created"); //$NON-NLS-1$
+            }
+        } catch (TmfAnalysisException | TmfTraceException e) {
+            module.dispose();
+            throw new TmfConfigurationException("Exception when setting trace", e); //$NON-NLS-1$
+        }
+        return module;
+    }
+
+    /**
+     * Removes configuration from trace:
+     *  - delete configuration file
+     *  - remove analysis module from trace object
+     *
+     * @param config
+     *          the configuration to remove
+     * @param trace
+     *          the
+     * @throws TmfConfigurationException if an error occurs
+     */
+    public static void remove(ITmfConfiguration config, @NonNull ITmfTrace trace) throws TmfConfigurationException {
+        IPath traceConfig = getTraceRootFolder(trace, config.getSourceTypeId());
+        traceConfig = traceConfig.append(File.separator).append(config.getId()).addFileExtension(TmfConfiguration.JSON_EXTENSION);
+        File configFile = traceConfig.toFile();
+        if ((!configFile.exists()) || !configFile.delete()) {
+            throw new TmfConfigurationException("InAndOut configuration file can't be deleted from trace: configId=" + config.getId()); //$NON-NLS-1$
+        }
+
+        // Remove and clear persistent data
+        try {
+            IAnalysisModule module = trace.removeAnalysisModule(ID + config.getId());
+            if (module != null) {
+                module.dispose();
+                module.clearPersistentData();
+            }
+        } catch (TmfTraceException e) {
+            throw new TmfConfigurationException("Error removing analysis module from trace: analysis ID=" + ID + config.getId(), e); //$NON-NLS-1$
+        }
+    }
+
+    @SuppressWarnings("null")
+    private static @NonNull IPath getTraceRootFolder(@NonNull ITmfTrace trace, String subFolder) {
+        String supplFolder = TmfTraceManager.getSupplementaryFileDir(trace);
+        IPath supplPath = new Path(supplFolder);
+        supplPath = supplPath.addTrailingSeparator().append(subFolder);
+        return supplPath;
+    }
+
+    /**
+     * Reads the configurations for a given trace
+     *
+     * @param trace
+     *            the trace to read configurations from
+     * @return list of configurations if any
+     * @throws TmfConfigurationException
+     *             if an error occurs
+     */
+    public static @NonNull List<ITmfConfiguration> readConfigurations(@NonNull ITmfTrace trace) throws TmfConfigurationException {
+        IPath rootPath = getTraceRootFolder(trace, SegmentSpecifierConfiguration.IN_AND_OUT_CONFIG_SOURCE_TYPE_ID);
+        File folder = rootPath.toFile();
+        List<ITmfConfiguration> list = new ArrayList<>();
+        if (folder.exists()) {
+            File[] listOfFiles = folder.listFiles();
+            for (File file : listOfFiles) {
+                IPath path = new Path(file.getName());
+                if (path.getFileExtension().equals(TmfConfiguration.JSON_EXTENSION)) {
+                    ITmfConfiguration config = TmfConfiguration.fromJsonFile(file);
+                    list.add(config);
+                }
+            }
+        }
+        return list;
+    }
+
 }
--- a/analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutDataProviderFactory.java
+++ b/analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutDataProviderFactory.java
@@ -1,0 +1,269 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Ericsson
+ *
+ * All rights reserved. This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0 which
+ * accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+
+package org.eclipse.tracecompass.incubator.internal.inandout.core.analysis;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+import org.eclipse.core.runtime.FileLocator;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Path;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.osgi.util.NLS;
+import org.eclipse.tracecompass.incubator.internal.inandout.core.Activator;
+import org.eclipse.tracecompass.tmf.core.config.ITmfConfiguration;
+import org.eclipse.tracecompass.tmf.core.config.ITmfConfigurationSourceType;
+import org.eclipse.tracecompass.tmf.core.config.ITmfDataProviderConfigurator;
+import org.eclipse.tracecompass.tmf.core.config.TmfConfiguration;
+import org.eclipse.tracecompass.tmf.core.config.TmfConfigurationSourceType;
+import org.eclipse.tracecompass.tmf.core.dataprovider.IDataProviderDescriptor;
+import org.eclipse.tracecompass.tmf.core.dataprovider.IDataProviderDescriptor.ProviderType;
+import org.eclipse.tracecompass.tmf.core.dataprovider.IDataProviderFactory;
+import org.eclipse.tracecompass.tmf.core.exceptions.TmfConfigurationException;
+import org.eclipse.tracecompass.tmf.core.model.DataProviderDescriptor;
+import org.eclipse.tracecompass.tmf.core.model.tree.ITmfTreeDataModel;
+import org.eclipse.tracecompass.tmf.core.model.tree.ITmfTreeDataProvider;
+import org.eclipse.tracecompass.tmf.core.signal.TmfSignalHandler;
+import org.eclipse.tracecompass.tmf.core.signal.TmfSignalManager;
+import org.eclipse.tracecompass.tmf.core.signal.TmfTraceClosedSignal;
+import org.eclipse.tracecompass.tmf.core.signal.TmfTraceOpenedSignal;
+import org.eclipse.tracecompass.tmf.core.trace.ITmfTrace;
+import org.eclipse.tracecompass.tmf.core.trace.TmfTraceManager;
+import org.eclipse.tracecompass.tmf.core.trace.experiment.TmfExperiment;
+import org.osgi.framework.Bundle;
+
+import com.google.common.collect.HashBasedTable;
+import com.google.common.collect.Table;
+
+/**
+ * Data provider factory for InAndOut analysis. It doesn't have any output but is able
+ * to create custom InAndOut analyses.
+ */
+@NonNullByDefault
+public class InAndOutDataProviderFactory implements IDataProviderFactory, ITmfDataProviderConfigurator {
+
+    /** Data Provider factory ID */
+    public static final String ID = "org.eclipse.tracecompass.incubator.inandout.core.analysis.inAndOutdataProviderFactory"; //$NON-NLS-1$
+
+    private static final ITmfConfigurationSourceType CONFIG_SOURCE_TYPE;
+    private static final String NAME = "In And Out Analysis"; //$NON-NLS-1$
+    private static final String DESCRIPTION = "Configure In And Out analysis using file description"; //$NON-NLS-1$
+
+    private static final String CONFIGURATOR_NAME = "InAndOut Configurator"; //$NON-NLS-1$
+    private static final String CONFIGURATOR_DESCRIPTION = "Configure custom InAndOut analysis"; //$NON-NLS-1$
+
+    private static final String CUSTOM_IN_AND_OUT_ANALYSIS_NAME = "InAndOut Analysis ({0})";  //$NON-NLS-1$
+    private static final String CUSTOM_IN_AND_OUT_ANALYSIS_DESCRIPTION = "Custom InAndOut analysis configured by \" {0}\""; //$NON-NLS-1$
+
+    private static final Table<String, ITmfTrace, ITmfConfiguration> fTmfConfigurationTable = HashBasedTable.create();
+
+    static {
+        Bundle bundle = Platform.getBundle(Activator.PLUGIN_ID);
+        IPath defaultPath = new Path(SegmentSpecifierConfiguration.IN_AND_OUT_CONFIG_SOURCE_SCHEMA);
+        URL url = FileLocator.find(bundle, defaultPath, null);
+        File schemaFile = null;
+        try {
+            schemaFile = new File(FileLocator.toFileURL(url).toURI());
+        } catch (URISyntaxException | IOException e) {
+            Activator.getInstance().logError("Failed to read schema file: " + SegmentSpecifierConfiguration.IN_AND_OUT_CONFIG_SOURCE_SCHEMA, e); //$NON-NLS-1$
+        }
+        CONFIG_SOURCE_TYPE = new TmfConfigurationSourceType.Builder()
+                .setId(SegmentSpecifierConfiguration.IN_AND_OUT_CONFIG_SOURCE_TYPE_ID)
+                .setDescription(DESCRIPTION)
+                .setName(NAME)
+                .setSchemaFile(schemaFile)
+                .build();
+    }
+
+    // Data provider descriptor meant to be used to create new InAndOut configurations and hence new data providers
+    private static final IDataProviderDescriptor DESCRIPTOR = new DataProviderDescriptor.Builder()
+            .setId(ID)
+            .setName(CONFIGURATOR_NAME)
+            .setDescription(CONFIGURATOR_DESCRIPTION)
+            .setProviderType(ProviderType.NONE)
+            .build();
+
+    /**
+     * Default constructor
+     */
+    public InAndOutDataProviderFactory() {
+        TmfSignalManager.register(this);
+    }
+
+    @Override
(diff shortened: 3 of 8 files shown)


analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/config/InAndOutConfigurationSource.java
bhufmann, line 1: Not needed... I'll remove it

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/config/JsonUtils.java
bhufmann, line 1: Not needed... I'll remove it

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/config/package-info.java
bhufmann, line 1: Not needed... I'll remove it

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/config/InAndOutConfigurationSource.java
bhufmann: Done

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/config/JsonUtils.java
bhufmann: Done

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/config/package-info.java
bhufmann: Done

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutAnalysisModule.java
```diff
@@ -50,23 +58,79 @@ public class InAndOutAnalysisModule extends InstrumentedCallStackAnalysis {
      */
     public static final String JSON = ".config.json"; //$NON-NLS-1$
 
+    private @Nullable ITmfConfiguration fConfiguartion;
```
PatrickTasse, line 61: fConfiguration

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutAnalysisModule.java
```diff
@@ -50,23 +58,79 @@ public class InAndOutAnalysisModule extends InstrumentedCallStackAnalysis {
...
+            return false;
+        }
+        try {
+            fConfiguartion = configuration;
+            String json = configuration.getJsonParameters();
+            if (json.equals("{}")) { //$NON-NLS-1$
+                throw new TmfConfigurationException("JSON parameter is not deserialize correctly"); //$NON-NLS-1$
+            }
+            SegmentSpecifierConfiguration config = SegmentSpecifierConfiguration.fromJsonString(json);
+            fSpecifiers = config.getSpecifiers();
```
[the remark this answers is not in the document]
bhufmann: For this example, it looks straight forward. In general, extension have to deserialize the JSON. If we pass an already deserialized JSON in map<String, Object> then extension needs to traverse the map (which could be map of maps>, instead of using deserialization using POJOs. This is a design choice. Not sure which one we should choose. 

If we go with the Map approach it would be consistent with the rest of cases in the server. Validation would have to be manual in the extension and we cannot use a validation library for schema validation unless the map is serialized back to JSON and fed to the validator.

If we go with the string, then many existing code needs to be deprecated to remove 2 ways of configuring. With a string, the extension can handle the JSON string as required, POJO, parameter map. But this would be different way of working and serialization/deserialization of json.

I need to think about it a bit more.

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutAnalysisModule.java
bhufmann: ok

abhinava-ericsson:
InAndOutAnalysis feels in conflict with the concept of a configurable Data Provider.
Ideally, if we were to apply the concept afresh, InAndOut would be a config type for each of the 4 data providers.

Or, if we were to support configuration at the experiment level, it would again be a config type, which could be applied to all analysis/ data providers supported by that experiment.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/model/views/ConfigurationQueryParameters.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    @Hidden
+    public @NonNull JsonNode getParameters() {
+        return parameters;
+    }
+
+    @SuppressWarnings("nls")
+    @Override
+    public String toString() {
+        return "QueryParameters [parameters=" + parameters + ", json=" + parameters.toString() + "]";
```
PatrickTasse, line 57: Both of these should print the same string?

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutAnalysisModule.java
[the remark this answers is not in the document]
PatrickTasse: Right now we have DataProviderService deserializing the parameters into a JsonNode (in ConfigurationQueryParameters class). Then it converts the JsonNode to string in ITmfDataProviderSource.createDataProviderDescriptors(). Later the analysis module deserializes the parameters string into a SegmentSpecifierConfiguration.

What if we had just left the DataProviderService use the default to deserialize the parameters into a Map. Then nothing would prevent the extension (the analysis module) to still be able to deserialize the parameters into a SegmentSpecifierConfiguration. It just has to get the json string from this at line 89: 

            String json = new Gson().toJson(configuration.getParameters());

So some extensions can optionally validate and deserialize the json string, when it's a complex object, and others can just use the map of parameters when it's a simple few values.

abhinava-ericsson:
> > InAndOutAnalysis feels to be in conflict with the concept of a configurable Data Provider. Ideally, if we were to apply the concept afresh, InAndOut would be a config type for each of the 4 data providers.
> 
> Why? A given data provider could configure the creation of one or more other data providers. It could even configure data providers with a different provider type. For example, the Linux ThreadStatusDataProvider can spawn a Critical Path data provider. The InAndOutDataProvider actually doesn't have a provider type that has a view type associated with. Its sole purpose is to configure and manage InAndOutAnalysis that spwans multiple views (flamechart, statistics, scatter chart etc.) with the same configuration.

My point was that all "InAndOut" does is that it configures the multiple analyses/ views (flamechart, statistics, scatter chart etc.). That's why there is no base InAndOutDataProvider.  So, it feels more like a ConfigurationType than an analyses. It can obviously be made to fit (as it has been in this patch).

> Are you suggesting going back to the original design? Now it would be the right time before the whole design gets merged.

No, I think the per DataProvider configuration is useful. But, later we can also support a per experiment configuration, which would configure all data providers the experiment supports. I said that because InAndOut analyses seems to do two things:
- Club multiple analyses together
- Configures each of those analyses

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/model/ConfigurationQueryParameters.java
```diff
@@ -23,9 +21,9 @@
 public interface ConfigurationQueryParameters {
 
     /**
-     * @return The query parameter map.
+     * @return The query parameters.
      */
     @NonNull
-    @Schema(required = true, description = "Parameters as specified in corresponding ConfigurationTypeDescriptor. Use key `path` for file URI string values.")
-    Map<String, Object> getParameters();
+    @Schema(required = true, description = "Parameters as JSON object as specified in the schema of the corresponding ConfigurationTypeDescriptor or as a Map<String, Object>. Use key `path` for file URI string values.")
```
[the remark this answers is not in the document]
bhufmann: Yes, this file is only used for swagger documentation.

An Object resolves to a JsonObject in generated swagger documentation as far as I understand. 

The client will have to put a map<String, Object> where the key is `parameters` and the value the actual input .JSON object for this configuration (see below for an example). I kept the parameters map because it is consistent to all the other endpoints. What do you suggest to do it differently? I mean in general in the implementation as well as for the swagger documention.

```json
{
    "parameters": {
        "name": "my InAndOut",
        "description" : "my description",
        "specifiers" : [
            {
            "label":"latency",
            "inRegex":"(\\S*)_entry",
            "outRegex":"(\\S*)_exit",
            "contextInRegex":"(\\S*)_entry",
            "contextOutRegex":"(\\S*)_exit",
            "classifier":"CPU"
            }
        ]
    },
    "typeId": "org.eclipse.tracecompass.incubator.internal.inandout.core.config"
}
```

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/model/views/ConfigurationQueryParameters.java
bhufmann: Yes, I need to update this. I had the class implemented differently and the toString was not updated.

bhufmann:
> > > InAndOutAnalysis feels to be in conflict with the concept of a configurable Data Provider. Ideally, if we were to apply the concept afresh, InAndOut would be a config type for each of the 4 data providers.
> > 
> > 
> > Why? A given data provider could configure the creation of one or more other data providers. It could even configure data providers with a different provider type. For example, the Linux ThreadStatusDataProvider can spawn a Critical Path data provider. The InAndOutDataProvider actually doesn't have a provider type that has a view type associated with. Its sole purpose is to configure and manage InAndOutAnalysis that spwans multiple views (flamechart, statistics, scatter chart etc.) with the same configuration.
> 
> My point was that all "InAndOut" does is that it configures the multiple analyses/ views (flamechart, statistics, scatter chart etc.). That's why there is no base InAndOutDataProvider. So, it feels more like a ConfigurationType than an analyses. It can obviously be made to fit (as it has been in this patch).
> 
> > Are you suggesting going back to the original design? Now it would be the right time before the whole design gets merged.
> 
> No, I think the per DataProvider configuration is useful. But, later we can also support a per experiment configuration, which would configure all data providers the experiment supports. I said that because InAndOut analyses seems to do two things:
> 
> * Club multiple analyses together
> * Configures each of those analyses

Only one analysis is configured that feeds multiple views, not multiple analyses.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/model/ConfigurationQueryParameters.java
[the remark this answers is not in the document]
PatrickTasse: I guess what I'm understanding is that at TSP protocol (swagger) level, we don't need to make the difference, the example you give fits both. So we could just describe it here by modifying the original to add 'as specified in the schema of...'. It's not clear to me how the sender would differentiate between sending a JSON object or a Map, or how it would look different over the TSP, so I'm not sure about describing it in the protocol as a choice.

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutDataProviderFactory.java
```diff
@@ -0,0 +1,244 @@
...
+        /*
+         * Note only one config type is available for this.
+         */
+        ITmfConfiguration defaultConfig = new TmfConfiguration.Builder()
+                .setId("Dummy ID") // not used //$NON-NLS-1$
+                .setSourceTypeId(typeId)
+                .setName("InAndOut config") //$NON-NLS-1$
+                .setDescription("Custom In And Out").build(); //$NON-NLS-1$
+
+        ITmfConfiguration config = TmfJsonConfiguration.fromJsonString(jsonParameters, defaultConfig);
```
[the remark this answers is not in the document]
PatrickTasse: My intuition would be to define the mandatory/common configuration properties (id, name, description) outside so that they could all be described in the TSP, while leaving the parameters object as completely opaque and only defined by the specific extension in its schema's 'properties' element.

analyses/org.eclipse.tracecompass.incubator.inandout.core/schema/in-and-out-analysis.json
```diff
@@ -0,0 +1,59 @@
...
+        },
+        "specifiers": {
+            "description": "array specifiers",
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "label": {
+                        "type": "string",
+                        "descritpion": "The label of the identifier"
```
abhinava-ericsson, line 24: A useful optional field here (apart from type and description) would be a list of values to select elements from. This could be used to provide an option to the user to select instead of type in.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/model/ConfigurationQueryParameters.java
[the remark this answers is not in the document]
bhufmann: A JSON object in swagger boils done to a map if it is deserialized as a map with default deserializer.

analyses/org.eclipse.tracecompass.incubator.inandout.core/schema/in-and-out-analysis.json
bhufmann: Yes, we have to follow the JSON schema specification to see what is possible (https://json-schema.org/). The designer of the schema and extension will have look into that. I know, that you, for example, can specify a regex pattern to limit the input.

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutDataProviderFactory.java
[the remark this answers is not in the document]
bhufmann: I had that intuition too initially, some implication on the implementation on how the input json is generated in the client and how it can be validated in the server (using a validation library) If we put it outside we will need to get a separate JSON Object to feed the json validator. If the name/description is part of the schema, that JSON object has to include it. If not, the schema can't have it. Both solutions are valid.

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutAnalysisModule.java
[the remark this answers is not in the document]
bhufmann: > Right now we have DataProviderService deserializing the parameters into a JsonNode (in ConfigurationQueryParameters class). Then it converts the JsonNode to string in ITmfDataProviderSource.createDataProviderDescriptors(). Later the analysis module deserializes the parameters string into a SegmentSpecifierConfiguration.
> 
> What if we had just left the DataProviderService use the default to deserialize the parameters into a Map. Then nothing would prevent the extension (the analysis module) to still be able to deserialize the parameters into a SegmentSpecifierConfiguration. It just has to get the json string from this at line 89:
> 
> ```
>         String json = new Gson().toJson(configuration.getParameters());
> ```
> 
> So some extensions can optionally validate and deserialize the json string, when it's a complex object, and others can just use the map of parameters when it's a simple few values.

I think you are right. We should pass a Map<String, Object> from the server towards ITmfDataProviderSource.createDataProviderDescriptors() (and ITmfConfigurationSource.create()) and not use string. This would make it consistent to other TSP endpoints, avoids multiple serialization/deserialization steps, avoids having to deprecate, manage 2 APIs during deprecation period and manage removal of old API (in global). Overall, the server API will be simplier.

Extension can use map lookups and if wanted convert it to a JSON string and feed it to a POJO and custom deserializer, as showed in your example. using Gson.

@abhinava-ericsson FYI, the API will use default Map<String, Object> for JSON objects. The pull requests will be updated for that.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/ConfigurationManagerService.java
```diff
@@ -155,11 +152,17 @@ public Response postConfiguration(@Parameter(description = CFG_TYPE_ID) @PathPar
...
+//            // Pass JSON object as json string
+//            params.put(TmfConfiguration.JSON_STRING_KEY, queryParameters.getParameters().toString());
+//        } else {
+//            // Convert it to a Map<String, Object>
+//            ObjectMapper mapper = new ObjectMapper();
+//            params = mapper.convertValue(queryParameters.getParameters(), new TypeReference<Map<String,Object>>(){});
+//        }
         try {
-            ITmfConfiguration config = configurationSource.create(params);
+            ITmfConfiguration config = configurationSource.create(queryParameters.getParameters().toString());
```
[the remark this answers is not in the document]
bhufmann: API will be changed to use Map<String, Object> instead of json string (see discussion for reasoning: https://github.com/eclipse-tracecompass-incubator/org.eclipse.tracecompass.incubator/pull/88#discussion_r1791829115)

bhufmann:
> > > > InAndOutAnalysis feels to be in conflict with the concept of a configurable Data Provider. Ideally, if we were to apply the concept afresh, InAndOut would be a config type for each of the 4 data providers.
> > > 
> > > 
> > > Why? A given data provider could configure the creation of one or more other data providers. It could even configure data providers with a different provider type. For example, the Linux ThreadStatusDataProvider can spawn a Critical Path data provider. The InAndOutDataProvider actually doesn't have a provider type that has a view type associated with. Its sole purpose is to configure and manage InAndOutAnalysis that spwans multiple views (flamechart, statistics, scatter chart etc.) with the same configuration.
> > 
> > 
> > My point was that all "InAndOut" does is that it configures the multiple analyses/ views (flamechart, statistics, scatter chart etc.). That's why there is no base InAndOutDataProvider. So, it feels more like a ConfigurationType than an analyses. It can obviously be made to fit (as it has been in this patch).
> > > Are you suggesting going back to the original design? Now it would be the right time before the whole design gets merged.
> > 
> > 
> > No, I think the per DataProvider configuration is useful. But, later we can also support a per experiment configuration, which would configure all data providers the experiment supports. I said that because InAndOut analyses seems to do two things:
> > 
> > * Club multiple analyses together
> > * Configures each of those analyses
> 
> Only one analysis is configured that feeds multiple views, not multiple analyses.

@abhinava-ericsson, @PatrickTasse 
I realized that for InAndOut analysis the remove method is not conceptionally correct and will be confusing for the end user. What mean is that there is one configuration that spawns multiple data providers (which is a valid use case), but when removing user would have to use of the list of data providers to remove the configuration and hence he rest of the data providers. For that use case we will need to have a separate interface (and endpoint) to remove the configuration which then removes the associated data providers. That would be much cleaner:
```
ITmfDataProviderSource.remove(ITmfTrace, String configID)
```

The endpoint could be as followed (or similar)

```
/experiments/{expUUID}/outputs/{outputId}/configs/{configId}
```

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
```diff
@@ -1129,6 +1148,247 @@ public Response getStyles(
...
+     * @param expUUID
+     *            desired experiment UUID
+     * @param outputId
+     *            Output ID for the data provider to query
+     * @param derivedOutputId
+     *            Output ID for the data provider to delete
+     * @return status and the deleted configuration instance, if successful
+     */
+    @DELETE
+    @Path("/{outputId}/{derivedOutputId}")
```
bhufmann, line 1336: Should we change the following? This would require that the derived output know how to delete it.
```
@Path("/{derivedOutputId}/")
```

As proposed earlier (https://github.com/eclipse-tracecompass-incubator/org.eclipse.tracecompass.incubator/pull/88#issuecomment-2399802653) to have a way to remove a config from the data provider instance which will remove the actual data provider(s) consequently.

@abhinava-ericsson @PatrickTasse WDYT?

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutDataProviderFactory.java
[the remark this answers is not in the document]
PatrickTasse: I'm ok with both solutions as long as it's clear in TSP what needs to be included in the map.

How about if the map contains 'name', 'id', 'description' and 'properties' where the value of properties is an object that matches the properties in the JSON schema? I think it would be less risky than having the application-specific elements at the same root level as the mandatory configuration elements, to avoid key name conflict.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
```diff
@@ -1138,4 +1170,94 @@ private static Response validateParameters(String outputId, QueryParameters quer
...
+     * @param queryParameters
+     *            the query parameters used to create a data provider
+     * @return a list of data provider descriptors
+     */
+    @POST
+    @Path("/{outputId}")
+    @Tag(name = OCG)
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(summary = "Get the list of outputs for this configuration", responses = {
```
PatrickTasse, line 1264: Needs update

abhinava-ericsson:
> > > > > InAndOutAnalysis feels to be in conflict with the concept of a configurable Data Provider. Ideally, if we were to apply the concept afresh, InAndOut would be a config type for each of the 4 data providers.
> > > > 
> > > > 
> > > > Why? A given data provider could configure the creation of one or more other data providers. It could even configure data providers with a different provider type. For example, the Linux ThreadStatusDataProvider can spawn a Critical Path data provider. The InAndOutDataProvider actually doesn't have a provider type that has a view type associated with. Its sole purpose is to configure and manage InAndOutAnalysis that spwans multiple views (flamechart, statistics, scatter chart etc.) with the same configuration.
> > > 
> > > 
> > > My point was that all "InAndOut" does is that it configures the multiple analyses/ views (flamechart, statistics, scatter chart etc.). That's why there is no base InAndOutDataProvider. So, it feels more like a ConfigurationType than an analyses. It can obviously be made to fit (as it has been in this patch).
> > > > Are you suggesting going back to the original design? Now it would be the right time before the whole design gets merged.
> > > 
> > > 
> > > No, I think the per DataProvider configuration is useful. But, later we can also support a per experiment configuration, which would configure all data providers the experiment supports. I said that because InAndOut analyses seems to do two things:
> > > 
> > > * Club multiple analyses together
> > > * Configures each of those analyses
> > 
> > 
> > Only one analysis is configured that feeds multiple views, not multiple analyses.
> 
> @abhinava-ericsson, @PatrickTasse I realized that for InAndOut analysis the remove method is not conceptionally correct and will be confusing for the end user. What mean is that there is one configuration that spawns multiple data providers (which is a valid use case), but when removing user would have to use of the list of data providers to remove the configuration and hence he rest of the data providers. For that use case we will need to have a separate interface (and endpoint) to remove the configuration which then removes the associated data providers. That would be much cleaner:
> 
> ```
> ITmfDataProviderSource.remove(ITmfTrace, String configID)
> ```
> 
> The endpoint could be as followed (or similar)
> 
> ```
> /experiments/{expUUID}/outputs/{outputId}/configs/{configId}
> ```

This would still be confusing for a user, because from their point of view, they configured and created a derived data provider (InAndOutDataProvider), with a configuration as a parameter. The configuration and other data providers were created as a side effect. And now instead of deleting that derived data provider, they are having to delete the configuration.

I still think that this is happening because:
1. we are treating InAndOut as an analysis instead of a config type. 
2. we are creating multiple data providers 

If we do want to do that, we should create a dummy derived InAndOutDataProvider, and delete it to:
1. delete all the accompanying data providers 
2. delete the config.

bhufmann:
> > > > > InAndOutAnalysis feels to be in conflict with the concept of a configurable Data Provider. Ideally, if we were to apply the concept afresh, InAndOut would be a config type for each of the 4 data providers.
> > > > 
> > > > 
> > > > Why? A given data provider could configure the creation of one or more other data providers. It could even configure data providers with a different provider type. For example, the Linux ThreadStatusDataProvider can spawn a Critical Path data provider. The InAndOutDataProvider actually doesn't have a provider type that has a view type associated with. Its sole purpose is to configure and manage InAndOutAnalysis that spwans multiple views (flamechart, statistics, scatter chart etc.) with the same configuration.
> > > 
> > > 
> > > My point was that all "InAndOut" does is that it configures the multiple analyses/ views (flamechart, statistics, scatter chart etc.). That's why there is no base InAndOutDataProvider. So, it feels more like a ConfigurationType than an analyses. It can obviously be made to fit (as it has been in this patch).
> > > > Are you suggesting going back to the original design? Now it would be the right time before the whole design gets merged.
> > > 
> > > 
> > > No, I think the per DataProvider configuration is useful. But, later we can also support a per experiment configuration, which would configure all data providers the experiment supports. I said that because InAndOut analyses seems to do two things:
> > > 
> > > * Club multiple analyses together
> > > * Configures each of those analyses
> > 
> > 
> > Only one analysis is configured that feeds multiple views, not multiple analyses.
> 
> @abhinava-ericsson, @PatrickTasse I realized that for InAndOut analysis the remove method is not conceptionally correct and will be confusing for the end user. What mean is that there is one configuration that spawns multiple data providers (which is a valid use case), but when removing user would have to use of the list of data providers to remove the configuration and hence he rest of the data providers. For that use case we will need to have a separate interface (and endpoint) to remove the configuration which then removes the associated data providers. That would be much cleaner:
> 
> ```
> ITmfDataProviderSource.remove(ITmfTrace, String configID)
> ```
> 
> The endpoint could be as followed (or similar)
> 
> ```
> /experiments/{expUUID}/outputs/{outputId}/configs/{configId}
> ```

@abhinava-ericsson I think there is a solution to make it clearer. @PatrickTasse suggested to return only one data provider descriptor when posting the config. This resulting data provider would be a derived data provider, which may have a graph associated with (e.g. timegraph) or has itself children data providers used for the InAndOut use case. The end user will always use the single derived data provider for removal. In case the derived data provider has children it will also remove the children.

With this we will always have a 1:1 relationship between configuration and derived data provider and the end-user will manage the derived data provider. We don't a separate code path and endpoint anymore as I earlier suggested and it will make things easier.

For the InAndOutAnalysis the hierarchy will look like the following:
```
--- InAndOut Configurator
    |--- Derived InAndOut (A)
        |--- InAndOut (A) - Flame Chart
        |--- InAndOut (A) - Latency Statistics 
        |--- InAndOut (A) - Latency Table
        |--- InAndOut (A) - Latency vs Time
    |--- Derived InAndOut (B)
        |--- InAndOut (B) - Flame Chart
        |--- InAndOut (B) - Latency Statistics 
        |--- InAndOut (B) - Latency Table
        |--- InAndOut (B) - Latency vs Time
```

The end-user will use the `InAndOut Configurator` to create a new InAndOut analysis (e.g.`InAndOut (A)`) and used `Derived InAndOut (A)` to remove it. As a nice side effect there will be automatic grouping and which can be visualized in the FE clients.

The API in `ITmfDataProviderSource` will look like this:

```Java
IDataProviderDescriptor create(String typeID, ITmfTrace trace, Map<String, Object> parameters);
void ITmfDataProviderSource.remove(ITmfTrace trace, IDataProviderDescriptor derivedDescriptor);
```
WDYT?

abhinava-ericsson:
> > > > > > InAndOutAnalysis feels to be in conflict with the concept of a configurable Data Provider. Ideally, if we were to apply the concept afresh, InAndOut would be a config type for each of the 4 data providers.
> > > > > 
> > > > > 
> > > > > Why? A given data provider could configure the creation of one or more other data providers. It could even configure data providers with a different provider type. For example, the Linux ThreadStatusDataProvider can spawn a Critical Path data provider. The InAndOutDataProvider actually doesn't have a provider type that has a view type associated with. Its sole purpose is to configure and manage InAndOutAnalysis that spwans multiple views (flamechart, statistics, scatter chart etc.) with the same configuration.
> > > > 
> > > > 
> > > > My point was that all "InAndOut" does is that it configures the multiple analyses/ views (flamechart, statistics, scatter chart etc.). That's why there is no base InAndOutDataProvider. So, it feels more like a ConfigurationType than an analyses. It can obviously be made to fit (as it has been in this patch).
> > > > > Are you suggesting going back to the original design? Now it would be the right time before the whole design gets merged.
> > > > 
> > > > 
> > > > No, I think the per DataProvider configuration is useful. But, later we can also support a per experiment configuration, which would configure all data providers the experiment supports. I said that because InAndOut analyses seems to do two things:
> > > > 
> > > > * Club multiple analyses together
> > > > * Configures each of those analyses
> > > 
> > > 
> > > Only one analysis is configured that feeds multiple views, not multiple analyses.
> > 
> > 
> > @abhinava-ericsson, @PatrickTasse I realized that for InAndOut analysis the remove method is not conceptionally correct and will be confusing for the end user. What mean is that there is one configuration that spawns multiple data providers (which is a valid use case), but when removing user would have to use of the list of data providers to remove the configuration and hence he rest of the data providers. For that use case we will need to have a separate interface (and endpoint) to remove the configuration which then removes the associated data providers. That would be much cleaner:
> > ```
> > ITmfDataProviderSource.remove(ITmfTrace, String configID)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The endpoint could be as followed (or similar)
> > ```
> > /experiments/{expUUID}/outputs/{outputId}/configs/{configId}
> > ```
> 
> @abhinava-ericsson I think there is a solution to make it clearer. @PatrickTasse suggested to return only one data provider descriptor when posting the config. This resulting data provider would be a derived data provider, which may have a graph associated with (e.g. timegraph) or has itself children data providers used for the InAndOut use case. The end user will always use the single derived data provider for removal. In case the derived data provider has children it will also remove the children.
> 
> With this we will always have a 1:1 relationship between configuration and derived data provider and the end-user will manage the derived data provider. We don't a separate code path and endpoint anymore as I earlier suggested and it will make things easier.
> 
> For the InAndOutAnalysis the hierarchy will look like the following:
> 
> ```
> --- InAndOut Configurator
>     |--- Derived InAndOut (A)
>         |--- InAndOut (A) - Flame Chart
>         |--- InAndOut (A) - Latency Statistics 
>         |--- InAndOut (A) - Latency Table
>         |--- InAndOut (A) - Latency vs Time
>     |--- Derived InAndOut (B)
>         |--- InAndOut (B) - Flame Chart
>         |--- InAndOut (B) - Latency Statistics 
>         |--- InAndOut (B) - Latency Table
>         |--- InAndOut (B) - Latency vs Time
> ```
> 
> The end-user will use the `InAndOut Configurator` to create a new InAndOut analysis (e.g.`InAndOut (A)`) and used `Derived InAndOut (A)` to remove it. As a nice side effect there will be automatic grouping and which can be visualized in the FE clients.
> 
> The API in `ITmfDataProviderSource` will look like this:
> 
> ```java
> IDataProviderDescriptor create(String typeID, ITmfTrace trace, Map<String, Object> parameters);
> void ITmfDataProviderSource.remove(ITmfTrace trace, IDataProviderDescriptor derivedDescriptor);
> ```
> 
> WDYT?

Yes. This definitely sounds like a nicer solution.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
abhinava-ericsson: I would say  `@Path("/{outputId}/{derivedOutputId}")` is cleaner. In general, it sounds like a nice idea to have all url references to derived output ids under a base output id.

abhinava-ericsson:
> > > > > > > InAndOutAnalysis feels to be in conflict with the concept of a configurable Data Provider. Ideally, if we were to apply the concept afresh, InAndOut would be a config type for each of the 4 data providers.
> > > > > > 
> > > > > > 
> > > > > > Why? A given data provider could configure the creation of one or more other data providers. It could even configure data providers with a different provider type. For example, the Linux ThreadStatusDataProvider can spawn a Critical Path data provider. The InAndOutDataProvider actually doesn't have a provider type that has a view type associated with. Its sole purpose is to configure and manage InAndOutAnalysis that spwans multiple views (flamechart, statistics, scatter chart etc.) with the same configuration.
> > > > > 
> > > > > 
> > > > > My point was that all "InAndOut" does is that it configures the multiple analyses/ views (flamechart, statistics, scatter chart etc.). That's why there is no base InAndOutDataProvider. So, it feels more like a ConfigurationType than an analyses. It can obviously be made to fit (as it has been in this patch).
> > > > > > Are you suggesting going back to the original design? Now it would be the right time before the whole design gets merged.
> > > > > 
> > > > > 
> > > > > No, I think the per DataProvider configuration is useful. But, later we can also support a per experiment configuration, which would configure all data providers the experiment supports. I said that because InAndOut analyses seems to do two things:
> > > > > 
> > > > > * Club multiple analyses together
> > > > > * Configures each of those analyses
> > > > 
> > > > 
> > > > Only one analysis is configured that feeds multiple views, not multiple analyses.
> > > 
> > > 
> > > @abhinava-ericsson, @PatrickTasse I realized that for InAndOut analysis the remove method is not conceptionally correct and will be confusing for the end user. What mean is that there is one configuration that spawns multiple data providers (which is a valid use case), but when removing user would have to use of the list of data providers to remove the configuration and hence he rest of the data providers. For that use case we will need to have a separate interface (and endpoint) to remove the configuration which then removes the associated data providers. That would be much cleaner:
> > > ```
> > > ITmfDataProviderSource.remove(ITmfTrace, String configID)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > The endpoint could be as followed (or similar)
> > > ```
> > > /experiments/{expUUID}/outputs/{outputId}/configs/{configId}
> > > ```
> > 
> > 
> > @abhinava-ericsson I think there is a solution to make it clearer. @PatrickTasse suggested to return only one data provider descriptor when posting the config. This resulting data provider would be a derived data provider, which may have a graph associated with (e.g. timegraph) or has itself children data providers used for the InAndOut use case. The end user will always use the single derived data provider for removal. In case the derived data provider has children it will also remove the children.
> > With this we will always have a 1:1 relationship between configuration and derived data provider and the end-user will manage the derived data provider. We don't a separate code path and endpoint anymore as I earlier suggested and it will make things easier.
> > For the InAndOutAnalysis the hierarchy will look like the following:
> > ```
> > --- InAndOut Configurator
> >     |--- Derived InAndOut (A)
> >         |--- InAndOut (A) - Flame Chart
> >         |--- InAndOut (A) - Latency Statistics 
> >         |--- InAndOut (A) - Latency Table
> >         |--- InAndOut (A) - Latency vs Time
> >     |--- Derived InAndOut (B)
> >         |--- InAndOut (B) - Flame Chart
> >         |--- InAndOut (B) - Latency Statistics 
> >         |--- InAndOut (B) - Latency Table
> >         |--- InAndOut (B) - Latency vs Time
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The end-user will use the `InAndOut Configurator` to create a new InAndOut analysis (e.g.`InAndOut (A)`) and used `Derived InAndOut (A)` to remove it. As a nice side effect there will be automatic grouping and which can be visualized in the FE clients.
> > The API in `ITmfDataProviderSource` will look like this:
> > ```java
> > IDataProviderDescriptor create(String typeID, ITmfTrace trace, Map<String, Object> parameters);
> > void ITmfDataProviderSource.remove(ITmfTrace trace, IDataProviderDescriptor derivedDescriptor);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > WDYT?
> 
> Yes. This definitely sounds like a nicer solution.

One thing I would say here is that in effect, we are introducing a notion of sub-dataproviders, which is slightly different than derived data providers. We'll need to handle it accordingly.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
bhufmann: Done

MatthewKhouzam (approved):

PR #90, opened 2024-10-02
[server] Stop creating a separate trace instance when posting a trace

siwei-zhang-work:
A separate trace instance was created to provide metadata info when a trace is queried. This commit removed the creation of that separate instance and open provide metadata info when the experiment is created.

### DIFF
--- a/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/Experiment.java
+++ b/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/Experiment.java
@@ -14,10 +14,13 @@ package org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.cor
 import java.io.Serializable;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
 import org.eclipse.core.resources.IResource;
+import org.eclipse.tracecompass.tmf.core.trace.ITmfTrace;
 import org.eclipse.tracecompass.tmf.core.trace.experiment.TmfExperiment;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
@@ -82,8 +85,11 @@ public final class Experiment implements Serializable {
      * @return the experiment model
      */
     public static Experiment from(TmfExperiment experiment, UUID expUUID) {
-        List<UUID> traceUUIDs = ExperimentManagerService.getTraceUUIDs(expUUID);
-        Set<Trace> traces = new LinkedHashSet<>(Lists.transform(traceUUIDs, uuid -> Trace.from(TraceManagerService.getOrCreateTraceInstance(uuid), uuid)));
+        Map<UUID, ITmfTrace> uuidToTraceInstances = ExperimentManagerService.getTraceInstances(expUUID);
+        Set<Trace> traces = uuidToTraceInstances.entrySet().stream()
+            .map(entry -> Trace.from(entry.getValue(), entry.getKey()))
+            .collect(Collectors.toCollection(LinkedHashSet::new));
+
         return new Experiment(experiment.getName(),
                 expUUID,
                 experiment.getNbEvents(),
--- a/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/ExperimentManagerService.java
+++ b/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/ExperimentManagerService.java
@@ -25,6 +25,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -94,6 +95,7 @@ import io.swagger.v3.oas.annotations.tags.Tag;
 public class ExperimentManagerService {
 
     private static final Map<UUID, List<UUID>> TRACE_UUIDS = Collections.synchronizedMap(new HashMap<>());
+    private static final Map<UUID, Map<UUID, ITmfTrace>> TRACE_INSTANCES = Collections.synchronizedMap(new HashMap<>());
     private static final Map<UUID, IResource> EXPERIMENT_RESOURCES = Collections.synchronizedMap(initExperimentResources());
     private static final Map<UUID, TmfExperiment> EXPERIMENTS = Collections.synchronizedMap(new HashMap<>());
     private static final Map<UUID, TraceAnnotationProvider> TRACE_ANNOTATION_PROVIDERS = Collections.synchronizedMap(new HashMap<>());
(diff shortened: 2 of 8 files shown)


trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/ExperimentManagerService.java
```diff
@@ -349,9 +350,15 @@ public Response postExperiment(@RequestBody(content = {
 
         // Instantiate the experiment and return it
         ITmfTrace[] traces = Lists.transform(traceUUIDs, uuid -> TraceManagerService.createTraceInstance(uuid)).toArray(new ITmfTrace[0]);
+        Map<UUID, ITmfTrace> uuidToTraceInstance = new HashMap<>();
+        for (UUID uuid : traceUUIDs) {
+            ITmfTrace trace = TraceManagerService.createTraceInstance(uuid);
+            uuidToTraceInstance.put(uuid, trace);
+        }
+        TRACE_INSTANCES.put(expUUID, uuidToTraceInstance);
```
bhufmann, line 358: This line should be only be done below when the actual TmfExperiment instantiation was successful (line 378)

bhufmann (changes requested):
Thanks for the update. I have some comments. Also, the unit tests are still failing. Please have look.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/ExperimentManagerService.java
```diff
@@ -349,9 +350,15 @@ public Response postExperiment(@RequestBody(content = {
 
         // Instantiate the experiment and return it
         ITmfTrace[] traces = Lists.transform(traceUUIDs, uuid -> TraceManagerService.createTraceInstance(uuid)).toArray(new ITmfTrace[0]);
+        Map<UUID, ITmfTrace> uuidToTraceInstance = new HashMap<>();
+        for (UUID uuid : traceUUIDs) {
+            ITmfTrace trace = TraceManagerService.createTraceInstance(uuid);
```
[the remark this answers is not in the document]
PatrickTasse: I think this can all be simplified:
At line 352 we have the ordered list of traceUUIDs that can later be obtained in Experiment.from(TmfExperiment, UUID) from getTraceUUIDs(expUUID).
The ordered list of created trace instances is fed to the experiment in initExperiment(). This list can also be later obtained in Experiment.from(TmfExperiment, UUID) from TmfExperiment.getTraces().
So if you iterate by index in both lists at the same time, you can access the corresponding values needed for Trace.from(ITmfTrace, UUID).

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/ExperimentManagerService.java
siwei-zhang-work: I didn't realize the traces have a guaranteed order, removed this part.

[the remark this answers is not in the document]
siwei-zhang-work: I didn't realize the traces have a guaranteed order, removed this part.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/ExperimentManagerService.java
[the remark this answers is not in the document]
siwei-zhang-work: Done.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/Experiment.java
```diff
@@ -82,8 +84,14 @@ public Experiment(@JsonProperty("name") String name,
      * @return the experiment model
      */
     public static Experiment from(TmfExperiment experiment, UUID expUUID) {
-        List<UUID> traceUUIDs = ExperimentManagerService.getTraceUUIDs(expUUID);
-        Set<Trace> traces = new LinkedHashSet<>(Lists.transform(traceUUIDs, uuid -> Trace.from(TraceManagerService.getOrCreateTraceInstance(uuid), uuid)));
+        Iterator<UUID> traceUUIDs = ExperimentManagerService.getTraceUUIDs(expUUID).iterator();
+        Iterator<ITmfTrace> traceInstances = experiment.getTraces().iterator();
```
bhufmann, line 88: The getTraces() returns an array of ITmfTrace. Each ITmfTrace might be another other TmfExperiment or trace with children. To get the actual traces you need to use TmfTraceManager.getTraceSet(experiment)) to get the leaf traces. However, TmfTraceManager.getTraceSet(experiment)) doesn't guarantee the same order. I fixed bug #51 where a UUID mismatch happened. This caused a problem in one of our non-open source downstream projects. 

Besides, relying that 2 collections return the same order is not ideal, because it can easily break by unrelated code changes. So, I think we'll need to be able to have a lookup from trace uuid to ITmfTrace similar to the first version of this PR.

bhufmann (changes requested):

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/Experiment.java
PatrickTasse: Good point. I agree. We can feed the TmfExperiment with traces in a certain order but we can't prevent it (or an subclass of TmfExperiment) from shuffling them around when returning getTraces().
Sorry for that extra back and forth, please revert to first version of PR while fixing the issues mentioned in earlier comments.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/ExperimentManagerService.java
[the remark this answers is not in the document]
PatrickTasse: Perhaps to address the first comment, build the uuidToTracesInstance map from traceUUIDs as the traces are being created, then after create the trace array from the map values. Would need to make sure to use a LinkedHashMap to preserve order.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/Experiment.java
siwei-zhang-work: Np, done.

bhufmann (changes requested):
It works well now. Please update the dispose method and it looks good from my side.

PatrickTasse (approved):

bhufmann (approved):
Looks good to me. Thanks!

PR #94, opened 2024-10-08
ros2: fix size of publisher/subscription GIDs

christophebedard:
This PR is related to (and is needed after) https://github.com/ros2/ros2_tracing/pull/138.

ROS 2 publishers and subscriptions have unique IDs called GIDs (global IDs), which is part of the ROS 2 middleware interface (`rmw`). Their values are based on the corresponding objects in the underlying middleware, usually the GUID (globally unique ID) of the DDS data writer and data reader.

Historically, the ROS 2 (`rmw`) GID was a 24-byte array, while the DDS GUIDs are usually 16-byte arrays, at least for the ["main" DDS implementations](https://docs.ros.org/en/rolling/Installation/DDS-Implementations.html) (Fast DDS, Cyclone DDS, Connext DDS). Tthe `rmw` GID was then just the DDS GUID plus some 8-byte (zero) padding.

Since https://github.com/ros2/rmw/pull/345, the `rmw` GID size has been reduced to 16 bytes, therefore matching the usual size of DDS GUIDs. However, ROS 2 traces were still incorrectly collecting 24-byte `rmw` GIDs: 16 valid bytes + 8 garbage bytes. In practice, these 8 extra bytes didn't change much, since nothing is currently relying on the GID values, at least not relying on getting the same GID for the same object from two different systems, where the first 16 bytes would be the same, but the last 8 bytes would most likely be different.

Since https://github.com/ros2/ros2_tracing/pull/138 fixed the ROS 2 tracing instrumentation to collect 16-bytes `rmw` GIDs, the trace processing code here needs to be adapted. Instead of removing the last 8 bytes to go from a 24-byte `rmw` GID to a 16-byte DDS GUIDs, just keep the first 16 bytes from the `rmw` GID. This will work for traces that still contain 24-byte `rmw` GIDs (since the last 8 bytes are garbage anyway) and will work for traces that contain 16-byte `rmw` GIDs.

### DIFF
--- a/tracetypes/org.eclipse.tracecompass.incubator.ros2.core/src/org/eclipse/tracecompass/incubator/internal/ros2/core/analysis/objects/Ros2ObjectsStateProvider.java
+++ b/tracetypes/org.eclipse.tracecompass.incubator.ros2.core/src/org/eclipse/tracecompass/incubator/internal/ros2/core/analysis/objects/Ros2ObjectsStateProvider.java
@@ -48,6 +48,8 @@ public class Ros2ObjectsStateProvider extends AbstractRos2StateProvider {
 
     private static final int VERSION_NUMBER = 0;
     private static final @NonNull String ROS_2_NAMESPACE_SEP = "/"; //$NON-NLS-1$
+    /** Size of an rmw GID, in practice */
+    private static final int RMW_GID_STORAGE_SIZE = 16;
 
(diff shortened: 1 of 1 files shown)


MatthewKhouzam (approved):
Yeah, looks great, I like the combined combineFastDdsGid usage.

PR #95, opened 2024-10-10
Cpp uftrace

MatthewKhouzam:


bhufmann (changes requested):
The commit message still has a change-id. Also, there is no PR description to explain what it does. The PR title should be updated as well.

Please update this PR accordingly.

MatthewKhouzam:
[vec.zip](https://github.com/user-attachments/files/17939297/vec.zip)
Here is a test. Careful with the zip. ;)

bhufmann (approved):
Looks good to me. Thanks!

PR #99, opened 2024-10-23
callstack: remove xml extension

arfio:
The extension was not removed in the previous patch which creates an error when launching Trace Compass. This commit removes the extension from the plugin.xml.


bhufmann (approved):
Looks good to me. Thanks!

PR #102, opened 2024-10-25
server: Add bookmarks endpoint support

kavehshahedi:
Using bookmarks, user can save custom data points where each corresponds to a specific time range. Each bookmark contains "name", "start", "end", and an optional "payload".

The following endpoints are introduced:
- [GET] Get all the bookmarks of an experiment
- [GET] Get a specific bookmark for an experiment
- [POST] Create a new bookmark for an experiment
- [PUT] Update an old bookmark of an experiment
- [DELETE] Delete a bookmark of an experiment

[Added] Corresponding endpoints for bookmarks in the trace server

### DIFF
--- a/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core.tests/src/org/eclipse/tracecompass/incubator/trace/server/jersey/rest/core/tests/services/BookmarkManagerServiceTest.java
+++ b/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core.tests/src/org/eclipse/tracecompass/incubator/trace/server/jersey/rest/core/tests/services/BookmarkManagerServiceTest.java
@@ -15,7 +15,10 @@ import static org.junit.Assert.*;
 
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
 
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.client.WebTarget;
@@ -26,6 +29,7 @@ import org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core
 import org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core.tests.stubs.BookmarkModelStub;
 import org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core.tests.stubs.ExperimentModelStub;
 import org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core.tests.utils.RestServerTest;
+import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
@@ -33,8 +37,8 @@ import org.junit.Test;
  * Test class for BookmarkManagerService
  *
  * @author Kaveh Shahedi
- * @since 10.1
  */
+@SuppressWarnings("null")
 public class BookmarkManagerServiceTest extends RestServerTest {
 
     private static final String BOOKMARK_NAME = "TEST";
@@ -43,37 +47,98 @@ public class BookmarkManagerServiceTest extends RestServerTest {
     private static final @NonNull BookmarkModelStub BOOKMARK = new BookmarkModelStub(BOOKMARK_NAME, START_TIME, END_TIME);
     private ExperimentModelStub experiment;
 
+    private static final String START = "start";
+    private static final String END = "end";
+
+    private static final String BOOKMARK_END_TIME_MATCH = "End time should match";
+    private static final String BOOKMARK_NAME_MATCH = "Bookmark name should match";
+    private static final String BOOKMARK_START_TIME_MATCH = "Start time should match";
+    private static final String NON_EXISTENT_BOOKMARK_STATUS_CODE = "Should return 404 for non-existent bookmark";
+    private static final String NON_EXISTENT_EXPERIMENT_STATUS_CODE = "Should return 404 for non-existent experiment";
+    private static final String NON_NULL_BOOKMARK = "Created bookmark should not be null";
+    private static final String NON_NULL_RESPONSE_BODY = "Response body should not be null";
+    private static final String NON_NULL_UUID = "UUID should not be null";
+    private static final String NON_NUMERIC_TIMES_STATUS_CODE = "Should return 400 for non-numeric times";
+    private static final String SUCCESSFUL_BOOKMARK_CREATION = "Bookmark creation should succeed";
+    private static final String SUCCESSFUL_STATUS_CODE = "Response status should be 200";
+
     /**
-     * Setup method to run before each test. Creates a clean experiment and removes all
-     * existing bookmarks.
+     * Setup method to run before each test
      */
     @Before
     public void setUp() {
         // Create the experiment first
         experiment = assertPostExperiment(CONTEXT_SWITCHES_UST_NOT_INITIALIZED_STUB.getName(),
-                                       CONTEXT_SWITCHES_UST_NOT_INITIALIZED_STUB);
+                CONTEXT_SWITCHES_UST_NOT_INITIALIZED_STUB);
         assertNotNull("Experiment should not be null", experiment);
-        assertNotNull("Experiment UUID should not be null", experiment.getUUID());
+        assertNotNull(NON_NULL_UUID, experiment.getUUID());
+    }
 
-        // Get all existing bookmarks and delete them
+    /**
+     * Tear down method to run after each test
+     */
+    @After
+    public void tearDown() {
+        // Remove all bookmarks
         WebTarget application = getApplicationEndpoint();
         WebTarget bookmarkTarget = application.path(EXPERIMENTS)
                 .path(experiment.getUUID().toString())
                 .path(BOOKMARKS);
 
-        Response response = bookmarkTarget.request().get();
-        assertEquals("GET request for bookmarks should return 200", 200, response.getStatus());
+        try (Response response = bookmarkTarget.request().get()) {
+            assertEquals("GET request for bookmarks should return 200", 200, response.getStatus());
 
-        if (response.getStatus() == 200) {
             BookmarkModelStub[] existingBookmarks = response.readEntity(BookmarkModelStub[].class);
             assertNotNull("Bookmark array should not be null", existingBookmarks);
 
             for (BookmarkModelStub bookmark : existingBookmarks) {
-                Response deleteResponse = bookmarkTarget.path(bookmark.getUUID().toString())
-                                                      .request()
-                                                      .delete();
-                assertEquals("DELETE request should return 200", 200, deleteResponse.getStatus());
+                try (Response deleteResponse = bookmarkTarget.path(bookmark.getUUID().toString())
+                        .request()
+                        .delete()) {
+                    assertEquals("DELETE request should return 200", 200, deleteResponse.getStatus());
+                }
             }
+        }
+    }
+
+    /**
+     * Test the bookmark endpoints with invalid experiment UUID (i.e.,
+     * non-existent).
+     */
+    @Test
+    public void testBookmarkEndpointsInvalidExperiment() {
+        WebTarget application = getApplicationEndpoint();
+        WebTarget bookmarkTarget = application.path(EXPERIMENTS)
+                .path(UUID.randomUUID().toString())
+                .path(BOOKMARKS);
+
+        // Test getting all bookmarks
+        try (Response response = bookmarkTarget.request().get()) {
+            assertEquals(NON_EXISTENT_EXPERIMENT_STATUS_CODE, 404, response.getStatus());
+        }
+
+        // Test getting a specific bookmark
+        try (Response response = bookmarkTarget.path(BOOKMARK.getUUID().toString()).request().get()) {
+            assertEquals(NON_EXISTENT_EXPERIMENT_STATUS_CODE, 404, response.getStatus());
+        }
+
+        // Test creating a bookmark
+        Map<String, Object> parameters = new HashMap<>();
+        parameters.put(NAME, BOOKMARK_NAME);
+        parameters.put(START, START_TIME);
+        parameters.put(END, END_TIME);
+        try (Response response = bookmarkTarget.request().post(Entity.json(new QueryParameters(parameters, Collections.emptyList())))) {
+            assertEquals(NON_EXISTENT_EXPERIMENT_STATUS_CODE, 404, response.getStatus());
+        }
+
+        // Test updating a bookmark
+        try (Response response = bookmarkTarget.path(BOOKMARK.getUUID().toString()).request().put(Entity.json(new QueryParameters(parameters, Collections.emptyList())))) {
+            assertEquals(NON_EXISTENT_EXPERIMENT_STATUS_CODE, 404, response.getStatus());
+        }
+
+        // Test deleting a bookmark
+        try (Response response = bookmarkTarget.path(BOOKMARK.getUUID().toString()).request().delete()) {
+            assertEquals(NON_EXISTENT_EXPERIMENT_STATUS_CODE, 404, response.getStatus());
         }
     }
 
@@ -90,27 +155,35 @@ public class BookmarkManagerServiceTest extends RestServerTest {
         // Test with null name
         Map<String, Object> parameters = new HashMap<>();
         parameters.put(NAME, null);
-        parameters.put("start", START_TIME);
-        parameters.put("end", END_TIME);
+        parameters.put(START, START_TIME);
+        parameters.put(END, END_TIME);
+        try (Response response = bookmarkTarget.request().post(Entity.json(new QueryParameters(parameters, Collections.emptyList())))) {
+            assertEquals("Should return 400 for null name", 400, response.getStatus());
 
-        Response response = bookmarkTarget.request().post(Entity.json(new QueryParameters(parameters, Collections.emptyList())));
-        assertEquals("Should return 400 for null name", 400, response.getStatus());
+        }
 
-        // Test with non-numeric start and end times
+        // Test with non-numeric start
         parameters.put(NAME, BOOKMARK_NAME);
-        parameters.put("start", "not a number");
-        parameters.put("end", "not a number");
+        parameters.put(START, "not a number");
+        try (Response response = bookmarkTarget.request().post(Entity.json(new QueryParameters(parameters, Collections.emptyList())))) {
+            assertEquals(NON_NUMERIC_TIMES_STATUS_CODE, 400, response.getStatus());
+        }
 
-        response = bookmarkTarget.request().post(Entity.json(new QueryParameters(parameters, Collections.emptyList())));
-        assertEquals("Should return 400 for non-numeric times", 400, response.getStatus());
+        // Test with non-numeric end
+        parameters.put(START, START_TIME);
+        parameters.put(END, "not a number");
+        try (Response response = bookmarkTarget.request().post(Entity.json(new QueryParameters(parameters, Collections.emptyList())))) {
+            assertEquals(NON_NUMERIC_TIMES_STATUS_CODE, 400, response.getStatus());
+        }
 
         // Test with end time before start time
         parameters.put(NAME, BOOKMARK_NAME);
-        parameters.put("start", END_TIME);
-        parameters.put("end", START_TIME);
+        parameters.put(START, END_TIME);
+        parameters.put(END, START_TIME);
+        try (Response response = bookmarkTarget.request().post(Entity.json(new QueryParameters(parameters, Collections.emptyList())))) {
+            assertEquals("Should return 400 for invalid time range", 400, response.getStatus());
+        }
 
-        response = bookmarkTarget.request().post(Entity.json(new QueryParameters(parameters, Collections.emptyList())));
-        assertEquals("Should return 400 for invalid time range", 400, response.getStatus());
     }
 
     /**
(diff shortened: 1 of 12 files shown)


trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/Bookmark.java
```diff
@@ -0,0 +1,136 @@
...
+     * @return the end time
+     */
+    public long getEnd() {
+        return fEnd;
+    }
+
+    /**
+     * Get the payload
+     *
+     * @return the JSON payload, empty JSON object if no payload was set
```
bhufmann, line 125: what is expected as payload? As a FE designer I would not know what to put there.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/BookmarkManagerService.java
```diff
@@ -0,0 +1,474 @@
...
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(summary = "Create a new bookmark in an experiment", responses = {
+            @ApiResponse(responseCode = "200", description = "Bookmark created successfully", content = @Content(schema = @Schema(implementation = Bookmark.class))),
+            @ApiResponse(responseCode = "400", description = INVALID_PARAMETERS, content = @Content(schema = @Schema(implementation = String.class))),
+            @ApiResponse(responseCode = "404", description = NO_SUCH_EXPERIMENT, content = @Content(schema = @Schema(implementation = String.class)))
+    })
+    public Response createBookmark(
+            @Parameter(description = EXP_UUID) @PathParam("expUUID") UUID expUUID,
+            @RequestBody(required = true) QueryParameters queryParameters) {
```
bhufmann, line 208: in the RequestBody you need to add a description and ```@Schema``` to specify the model implementation `BookmarkQueryParameters`. Otherwise the swagger doc generation will references `QueryParameter`, but it doesn't exist. You can also add an ```@ExampleObject``` to provide an example.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/model/BookmarkQueryParameters.java
```diff
@@ -0,0 +1,40 @@
...
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.media.Schema.RequiredMode;
+
+/**
+ * Parameters for bookmark creation and update operations
+ *
+ * @author Kaveh Shahedi
+ * @since 10.1
+ */
+public interface BookmarkQueryParameters {
```
bhufmann, line 22: This doesn't reflect the actual implementation. The implemenation uses QueryParameters which returns the parameter map. Your implementation uses the parameter map. Alternative you could have a implementation that pre-parses the JSON received over the network and hence you can have get the name, start and end directly without parsing the parameters map. In most cases in the server, the parameter map is used. The ConfigurationManagerService is done a bit differently, where some default TSP parameters (e.g. name or description) are defined and parsed. 

Besides, I don't see the `value` parameter anywhere.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/Bookmark.java
```diff
@@ -0,0 +1,136 @@
+package org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core.services;
```
bhufmann, line 1: Class misses the copyright header

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/BookmarkManagerService.java
```diff
@@ -0,0 +1,474 @@
...
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(summary = "Update an existing bookmark in an experiment", responses = {
+            @ApiResponse(responseCode = "200", description = "Bookmark updated successfully", content = @Content(schema = @Schema(implementation = Bookmark.class))),
+            @ApiResponse(responseCode = "400", description = INVALID_PARAMETERS, content = @Content(schema = @Schema(implementation = String.class))),
+            @ApiResponse(responseCode = "404", description = "Experiment or bookmark not found", content = @Content(schema = @Schema(implementation = String.class)))
+    })
+    public Response updateBookmark(
+            @Parameter(description = EXP_UUID) @PathParam("expUUID") UUID expUUID,
+            @Parameter(description = "Bookmark UUID") @PathParam("bookmarkUUID") UUID bookmarkUUID,
+            @RequestBody(required = true) QueryParameters queryParameters) {
```
bhufmann, line 317: Comment from POST applies here too.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/Bookmark.java
```diff
@@ -0,0 +1,136 @@
...
+    private static final long serialVersionUID = -3626414315455912960L;
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
+    private final UUID fUUID;
+    private final String fName;
+    private final String fExperimentId;
+    private final long fStart;
+    private final long fEnd;
+    private final JsonNode fPayload;
+
```
bhufmann, line 26: In Trace Compass a bookmark could have a color (RGBA). I wonder if we should add that right away. We could add `OutputElementStyle` object here. The FE will then translate the color to `OutputElementStyle`. Or we just keep a single color value.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/BookmarkManagerService.java
```diff
@@ -0,0 +1,474 @@
...
+        if (errorMessage != null) {
+            return Response.status(Status.BAD_REQUEST).entity(errorMessage).build();
+        }
+
+        TmfExperiment experiment = ExperimentManagerService.getExperimentByUUID(expUUID);
+        if (experiment == null) {
+            return Response.status(Status.NOT_FOUND).entity(NO_SUCH_EXPERIMENT).build();
+        }
+
+        String name = Objects.requireNonNull((String) parameters.get("name")); //$NON-NLS-1$
```
bhufmann, line 225: This causes a NullPointerException. The server will return an error 500 internal error, but it won't help the FE user to know what the problem. You need to check for null and then response with a proper error (400) and error string.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/BookmarkManagerService.java
```diff
@@ -0,0 +1,474 @@
...
+        long start = Objects.requireNonNull((Number) parameters.get("start")).longValue(); //$NON-NLS-1$
```
bhufmann, line 226: same here

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/BookmarkManagerService.java
```diff
@@ -0,0 +1,474 @@
...
+        long end = Objects.requireNonNull((Number) parameters.get("end")).longValue(); //$NON-NLS-1$
```
bhufmann, line 227: same here

bhufmann (changes requested):
First review. I haven't tested the functionality, I only tried to generate the openapi.yaml. The resulting file has errors and Swagger preview in vscode can't parse it (see comment at POST/PUT). 

I also, have some initial comments. 

Unit tests should be added as well to be able to test the functionality.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/Bookmark.java
kavehshahedi: This comes back to the requirements of Trace-Server/TraceCompass. This field is supposed to contain some metadata of other types of information that might be required to be saved in a bookmark. Since the fields of bookmark were not clear to me, I did put this extra field to be more flexible.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/BookmarkManagerService.java
kavehshahedi: I guess this issue has been already handled here:
```
Map<String, Object> parameters = queryParameters.getParameters();
String errorMessage = QueryParametersUtil.validateBookmarkQueryParameters(parameters);
if (errorMessage != null) {
    return Response.status(Status.BAD_REQUEST).entity(errorMessage).build();
}
```

So if these variables are null, the validator will return the corresponding error code for FE procedures.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/Bookmark.java
bhufmann: The point of the TSP is that the clients (front-end) can make sense of the data. If a server is returning arbitrary data, then a generic client implementation won't know how to input the data or display the data. Otherwise, we need custom client implementation for each custom bookmark data. 

In the application, we can have multiple types of bookmarks, which have common fields and bookmark type specific fields. What you're introducing I would call global bookmark, which has a start, end (optional), name, description. It may have a colour associated with (OutputElementStyle). This is similar to lost event markers. 

In the future, and it's not in scope for the initial bookmark support, we might create bookmarks on specific views, series, rows etc. which will have additional information, that needs to be stored in the bookmark.

What is in scope, however, is that the data structures that we define know can be augmented to support other use cases.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/model/BookmarkQueryParameters.java
kavehshahedi: I investigated this comment, but I couldn't figure out what should be done. I followed the same structure that is present in `ExperimentManagerService` class. Specifically:
```java
public Response postExperiment(@RequestBody(content = {
            @Content(schema = @Schema(implementation = ExperimentQueryParameters.class))
    }, required = true) QueryParameters queryParameters)
```
As it is implemented here, the `postExperiment` is using `ExperimentQueryParameters` schema, but also receiving a `QueryParameters` as argument. 

Also, I didn't understand the `value` parameter issue. Can you explain it a bit more?

P.S.: I ran the swagger yaml generation process, and it was successful.

bhufmann: It looks like `ExperimentQueryParameters` swagger schema is incorrect and doesn't match the actual implementation. I should have bee defined similarly to `RequestedQueryParameters`.

The implementation for experiments ask for the following input:
```
{
    parameters: {
       "name": "experiment name",
       "traces: ["uuid1", "uuid2"]
    }
}
```

but the actual swagger documentation says:

```
{
     "name": "experiment name",
     "traces: ["uuid1", "uuid2"]
}
```

So, if you'd like to use `parameters` key as you have implemented in the `BookmarkServiceManager`, then swagger documentation needs to be changed. Alternatively, you adjust the implementation to match the documentation. Maybe the first option is better, to be consistent with other endpoints.

About the comment about `value`, I'm not sure anymore, maybe it was about the `payload` that is part of the query parameters but there is no documentation.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core.tests/src/org/eclipse/tracecompass/incubator/trace/server/jersey/rest/core/tests/services/BookmarkManagerServiceTest.java
bhufmann, line 1: Excellent tests. It's increases the quality of the code and helps for testing and reviewing the source code.

bhufmann (changes requested):
Thanks for the contribution. Code looks good and I appreciate all the details and unit tests. There are few comments that I added.

One more thing, is that the bookmarks implementation via Trace Server is totally separate from the bookmark implementation that is available in classic Trace Compass and hence there is no common ground for reuse. I'm still debating (with myself) if we should align the implementation or if we continue with a separate implementation. I'm aware that the Trace Server implementation and how it's integrate in the clients differs in certain areas from the classic Trace Compass already and it's getting more and more (project model, configuration). One reason for it that certain features in classic Trace Compass have not decoupled UI and core implementation. One big one is the project model (workspace model), which it hard to re-create certain features in the trace server and its clients without re-implementing or painful restructuring in Trace Compass classic.  I'll ask other committers to weigh in here so that we find a way forward.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/model/Bookmark.java
```diff
@@ -0,0 +1,63 @@
...
+    @NonNull
+    @Schema(description = "User defined name for the bookmark")
+    String getName();
+
+    /**
+     * @return The experiment ID.
+     */
+    @NonNull
+    @Schema(description = "The experiment's unique identifier this bookmark belongs to")
+    String getExperimentId();
```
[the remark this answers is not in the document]
kavehshahedi: You are 100% correct. This property should've been deleted, but I forgot to do it. I'll fix it. Thanks for the notice.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/Bookmark.java
```diff
@@ -0,0 +1,136 @@
...
+ * @author Kaveh Shahedi
+ * @since 10.1
+ */
+public class Bookmark implements Serializable {
+    private static final long serialVersionUID = -3626414315455912960L;
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
+    private final UUID fUUID;
+    private final String fName;
+    private final String fExperimentId;
```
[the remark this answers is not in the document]
kavehshahedi: Yes, this is not necessary anymore since it was supposed to be removed in the previous commit. Will be fixed!

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/BookmarkManagerService.java
```diff
@@ -0,0 +1,474 @@
...
+     * Retrieve all bookmarks for a specific experiment
+     *
+     * @param expUUID
+     *            UUID of the experiment
+     * @return Response containing the list of bookmarks
+     */
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(summary = "Get all bookmarks for an experiment", responses = {
+            @ApiResponse(responseCode = "200", description = "Returns the list of bookmarks", content = @Content(array = @ArraySchema(schema = @Schema(implementation = Bookmark.class)))),
```
PatrickTasse, line 141: For Swagger should be Bookmark from model package.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/BookmarkManagerService.java
```diff
@@ -0,0 +1,474 @@
...
+     *            UUID of the experiment
+     * @param bookmarkUUID
+     *            UUID of the bookmark to retrieve
+     * @return Response containing the bookmark
+     */
+    @GET
+    @Path("/{bookmarkUUID}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(summary = "Get a specific bookmark from an experiment", responses = {
+            @ApiResponse(responseCode = "200", description = "Returns the bookmark", content = @Content(schema = @Schema(implementation = Bookmark.class))),
```
PatrickTasse, line 169: For Swagger should be Bookmark from model package.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/BookmarkManagerService.java
```diff
@@ -0,0 +1,474 @@
...
+     *            UUID of the experiment
+     * @param queryParameters
+     *            Parameters for creating the bookmark
+     * @return Response containing the created bookmark
+     */
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(summary = "Create a new bookmark in an experiment", responses = {
+            @ApiResponse(responseCode = "200", description = "Bookmark created successfully", content = @Content(schema = @Schema(implementation = Bookmark.class))),
```
PatrickTasse, line 202: For Swagger should be Bookmark from model package.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/BookmarkManagerService.java
```diff
@@ -0,0 +1,474 @@
...
+     * @param queryParameters
+     *            Parameters for updating the bookmark
+     * @return Response containing the updated bookmark
+     */
+    @PUT
+    @Path("/{bookmarkUUID}")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(summary = "Update an existing bookmark in an experiment", responses = {
+            @ApiResponse(responseCode = "200", description = "Bookmark updated successfully", content = @Content(schema = @Schema(implementation = Bookmark.class))),
```
PatrickTasse, line 297: For Swagger should be Bookmark from model package.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/BookmarkManagerService.java
```diff
@@ -0,0 +1,474 @@
...
+     *            UUID of the experiment
+     * @param bookmarkUUID
+     *            UUID of the bookmark to delete
+     * @return Response containing the deleted bookmark
+     */
+    @DELETE
+    @Path("/{bookmarkUUID}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(summary = "Delete a bookmark from an experiment", responses = {
+            @ApiResponse(responseCode = "200", description = "Bookmark deleted successfully", content = @Content(schema = @Schema(implementation = Bookmark.class))),
```
PatrickTasse, line 375: For Swagger should be Bookmark from model package.

PatrickTasse (changes requested):

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/BookmarkManagerService.java
```diff
@@ -0,0 +1,421 @@
...
+            return Response.status(Status.NOT_FOUND).entity(EndpointConstants.BOOKMARK_NOT_FOUND).build();
+        }
+
+        try {
+            IMarker[] markers = findBookmarkMarkers(editorFile);
+            IMarker marker = findMarkerByUUID(markers, bookmarkUUID);
+            if (marker == null) {
+                return Response.status(Status.NOT_FOUND).entity(EndpointConstants.BOOKMARK_NOT_FOUND).build();
+            }
+            return Response.ok(markerToBookmark(marker)).build();
```
PatrickTasse, line 149: Should null-check the bookmark for safety.

PatrickTasse (dismissed):

bhufmann (changes requested):
Just some formal comments and test cases. Add a test case that validates the error if a experiment doesn't exist on the server (i.e. UUID not available). For that you can have on test method that calls each endpoint with a UUID that doesn't exists on the server. 

Otherwise it' looks good to me.

bhufmann (dismissed):
Looks good to me. Thank your this contribution.

bhufmann (approved):

bhufmann (changes requested):
The Bookmark tag needs to be added to the TraceServerOpenApiResource, like the others. Otherwise the generated yaml file doesn't have a description for Bookmarks.

bhufmann (approved):
Looks good to me. Thanks!

MatthewKhouzam (commented):
do you want to squash or reflow these commits? I am not thrilled with the corrections in the pr being not squashed.

MatthewKhouzam (approved):
LGTM

MatthewKhouzam (commented):
Looks great! (orphaned commend. Github skill issue on my side.)

PR #103, opened 2024-10-29
Swagger updates for global and output configuration

bhufmann:
- Update documentation of Configuration and ConfigurationQueryParameters. 
Add name and description ConfigurationQueryParameters used for global configuration.
- Add configuration service per output (data provider)
This addition enables clients to create derived data providers from an existing data provider. This change also includes updates the "DataProvider" descriptor to include optional parent ID and configuration object to be included in derived data providers.

### DIFF
--- a/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/ConfigurationManagerService.java
+++ b/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/ConfigurationManagerService.java
@@ -142,8 +142,7 @@ public class ConfigurationManagerService {
     })
     public Response postConfiguration(@Parameter(description = CFG_TYPE_ID) @PathParam("typeId") String typeId,
             @RequestBody(description = CFG_CREATE_DESC + " " + CFG_KEYS_DESC, content = {
-                    @Content(examples = @ExampleObject("{\"parameters\":{" + CFG_PATH_EX +
-                            "}}"), schema = @Schema(implementation = org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core.model.ConfigurationQueryParameters.class))
+                    @Content(examples = @ExampleObject(CFG_PATH_EX), schema = @Schema(implementation = org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core.model.ConfigurationQueryParameters.class))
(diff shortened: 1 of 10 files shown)


trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/ConfigurationManagerService.java
```diff
@@ -225,7 +225,7 @@ public Response putConfiguration(@Parameter(description = CFG_TYPE_ID) @PathPara
             @RequestBody(description = CFG_UPDATE_DESC + " " + CFG_KEYS_DESC, content = {
                     @Content(examples = @ExampleObject("{\"parameters\":{" + CFG_PATH_EX +
                             "}}"), schema = @Schema(implementation = org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core.model.ConfigurationQueryParameters.class))
-            }, required = true) ConfigurationQueryParameters queryParameters) {
+            }, required = true) InputConfigurationQueryParameters queryParameters) {
```
PatrickTasse, line 228: Maybe there is an issue with Swagger generation that would prevent this that I don't understand, but would it be possible to have the same concept in the views package with views.ConfigurationQueryParameters here without typeId and views.OutputConfigurationParameters that extends it to add typeId?

bhufmann: I updated it as you suggested, plus changed the swagger doc generation to  use `allOf` schema tag for inheritance. It seems to work and it's cleaner.

PatrickTasse (changes requested):
The example object in ConfigurationManagerService (POST and PUT) is missing 'name' which is mandatory, and perhaps 'description' which is optional.
I noticed that in DataProviderService we don't provide any example object and the example is automatically generated with all parameters with default values. Perhaps this one here was made explicit to provide "path" as an example?

bhufmann:
> The example object in ConfigurationManagerService (POST and PUT) is missing 'name' which is mandatory, and perhaps 'description' which is optional. I noticed that in DataProviderService we don't provide any example object and the example is automatically generated with all parameters with default values. Perhaps this one here was made explicit to provide "path" as an example?

I can update the example. 

For the dp configuration, I didn't put an explicit example, because an example is generated. Generated examples has less details, though.

PatrickTasse (approved):

PR #115, opened 2024-11-08
Align OutputConfigurationQueryParameter/Configuration for sourceTypeId

bhufmann:
In the OutputConfigurationQueryParameter the configuration source type id was called typeId and it creates a Configuration object where the same field is called sourceTypeId for the same field. When serializing the Configuration in the derived DataProviderDescriptor over TSP the field is called sourceTypeId and this is not consistent. The returned Configuration should have the exact field names that are used when creating the data provider and the respective configuration.

The Configuration is already API while OutputConfigurationQueryParameter is not, and it's better to change the name there.

### DIFF
--- a/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/model/views/OutputConfigurationQueryParameters.java
+++ b/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/model/views/OutputConfigurationQueryParameters.java
@@ -20,7 +20,7 @@ import org.eclipse.tracecompass.tmf.core.config.TmfConfiguration;
  * Definition of a parameters object received by the server from a client for configurations.
  */
 public class OutputConfigurationQueryParameters extends ConfigurationQueryParameters {
-    private @NonNull String typeId;
+    private @NonNull String sourceTypeId;
 
     /**
      * Constructor for Jackson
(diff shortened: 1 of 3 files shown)


trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/model/views/OutputConfigurationQueryParameters.java
```diff
@@ -38,28 +38,28 @@ public OutputConfigurationQueryParameters() {
...
      *            the description of the configuration
-     * @param typeId
+     * @param souruceTypeId
      *            the typeId of the configuration
      *
      * @param parameters
      *            Map of parameters
      */
-    public OutputConfigurationQueryParameters(String name, String description, String typeId, Map<String, Object> parameters) {
+    public OutputConfigurationQueryParameters(String name, String description, String souruceTypeId, Map<String, Object> parameters) {
```
PatrickTasse, line 47: sourceTypeId

PatrickTasse (changes requested):

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/model/views/OutputConfigurationQueryParameters.java
bhufmann: Done. Thanks

PatrickTasse (approved):

PR #122, opened 2024-11-19
Fix trace server version and display

### LINKED ISSUE #121: trace-server: Identifier service shows version incorrectly
The identifier version is shown as `major.patch.minor` instead of `major.minor.patch`. This needs to be fixed

bhufmann:
- Update script to match the rest.core version to the release version
- Match rest.core plug-in version with release version
- Fix version display provided by IdentifierService

Note: The rest.core plug-in version is retrieved to show in the identifier service. This is the only way to retrieve the version programmatically and hence the plug-in version and release version has to match.

fixes #121


MatthewKhouzam (approved):
Good catch for versioning.

PR #127, opened 2024-11-24
ros2: support client/service instrumentation

christophebedard:
This adds support for the new client/service (i.e., RPC) instrumentation in ROS 2, see https://github.com/ros2/ros2_tracing/pull/145.

1. In the objects analysis, create client and service objects
2. In the messages analysis, create the following instances:
    1. Request publication
    2. Request take and callback
    3. Response publication
    4. Response take
    5. Message transport for requests and responses
3. In the messages dataprovider, display the above instances

There is one limitation. Normal message publications and message takes have instrumentation that provides a "start time." For example, for message publications, the `ros2:rclcpp_publish` tracepoint is the start and the `ros2:rmw_publish` tracepoint is the end of a message publication. This allows us to attribute a duration to the publication and therefore display a time graph state. However, we only have a single tracepoint for client/service-related publication/take instances, so we do not have any duration data. For now, just hardcode a 5000 ns duration so that time graph states are visible enough.

---

Example:  
Screenshot from 2024-11-24 11-53-36  
(trace: [2024-11-23__test_tracetools__test_service_pingpong__service-e2e.zip](https://github.com/user-attachments/files/17895029/2024-11-23__test_tracetools__test_service_pingpong__service-e2e.zip))

Clients (S↗️) and services (S↘️) are displayed under nodes (🔲) on the left. Chronological explanation:

1. The `/ping` client under the `/test_service_ping` node sends a request (short-ish time graph state from which the outgoing arrow starts)
2. The `/ping` service under the `/test_service_pong` node receives (takes) the request (thinner time graph state that the incoming arrow points to, right before the long time graph state)
3. The `/ping` service calls executes a callback for the request (long time graph state)
4. The `/ping` service sends a response inside the request callback (time at which the outgoing arrow starts)
5. The `/ping` client receives (takes) the request (thinner time graph state that the incoming arrow points to)

### DIFF
--- a/tracetypes/org.eclipse.tracecompass.incubator.ros2.core/src/org/eclipse/tracecompass/incubator/internal/ros2/core/Activator.java
+++ b/tracetypes/org.eclipse.tracecompass.incubator.ros2.core/src/org/eclipse/tracecompass/incubator/internal/ros2/core/Activator.java
@@ -22,9 +22,11 @@ import org.eclipse.tracecompass.incubator.internal.ros2.core.model.messages.Ros2
 import org.eclipse.tracecompass.incubator.internal.ros2.core.model.messages.Ros2TakeInstance;
 import org.eclipse.tracecompass.incubator.internal.ros2.core.model.messages.Ros2TimerCallbackInstance;
 import org.eclipse.tracecompass.incubator.internal.ros2.core.model.objects.Ros2CallbackObject;
+import org.eclipse.tracecompass.incubator.internal.ros2.core.model.objects.Ros2ClientObject;
 import org.eclipse.tracecompass.incubator.internal.ros2.core.model.objects.Ros2NodeObject;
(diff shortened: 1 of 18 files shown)


christophebedard:
I have to say that the ROS 2 plugin (especially the main view: the messages view) could use some serious refactoring/improvements, even before this PR. I'll try to find some time to do some cleanup.

MatthewKhouzam (approved):
Looks good,

MatthewKhouzam:
Uh... @christophebedard  want to merge this? You're a committer after all :)

christophebedard:
I was waiting for the new ROS 2 instrumentation to be reviewed just in case, but it's taking a bit longer than anticipated. Hopefully this week!

christophebedard:
Alright, the new instrumentation has been merged into ROS 2, so I'm merging this!

PR #130, opened 2024-11-28
uftrace: demangle c++ names

MatthewKhouzam:
Use CDT's cppFilt to demange c++ names if available.


bhufmann (approved):
Looks good to me. Thanks for cherry-picking it to this branch.

PR #133, opened 2024-12-04
Trace event stable diffusion

MatthewKhouzam:
* Handle if the "traceEvents" tag is lower in the file.
* Handle if the TID is not an integer. It appears to be out of spec, but stable diffusion does it that way.

### DIFF
--- a/tracetypes/org.eclipse.tracecompass.incubator.traceevent.core/src/org/eclipse/tracecompass/incubator/internal/traceevent/core/event/TraceEventAspects.java
+++ b/tracetypes/org.eclipse.tracecompass.incubator.traceevent.core/src/org/eclipse/tracecompass/incubator/internal/traceevent/core/event/TraceEventAspects.java
@@ -159,7 +159,14 @@ public class TraceEventAspects {
         @Override
         public @Nullable Integer resolve(@NonNull ITmfEvent event) {
             if (event instanceof TraceEventEvent) {
-                return ((TraceEventEvent) event).getField().getTid();
+                Object tid = ((TraceEventEvent) event).getField().getTid();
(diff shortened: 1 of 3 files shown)


tracetypes/org.eclipse.tracecompass.incubator.traceevent.core/src/org/eclipse/tracecompass/incubator/internal/traceevent/core/event/TraceEventAspects.java
```diff
@@ -159,7 +159,14 @@ public boolean isHiddenByDefault() {
         @Override
         public @Nullable Integer resolve(@NonNull ITmfEvent event) {
             if (event instanceof TraceEventEvent) {
-                return ((TraceEventEvent) event).getField().getTid();
+                Object tid = ((TraceEventEvent) event).getField().getTid();
+                if (tid instanceof Integer) {
+                    return (Integer) tid;
+                }
+                if (tid != null) {
```
bhufmann, line 166: Not knowing the traces, in this case is tid a name or a number string? if it is a number string you could do parseInt() or something like that.

Add a comment why it's done like that.

bhufmann (changes requested):

tracetypes/org.eclipse.tracecompass.incubator.traceevent.core/src/org/eclipse/tracecompass/incubator/internal/traceevent/core/event/TraceEventAspects.java
MatthewKhouzam: I think the new way is better.

PR #147, opened 2025-01-28
Configurable reports data provider

kavehshahedi:
### What it does
This PR introduces a new Reports Data Provider feature that enables users to:
- Define different types of reports (image, text, HTML)
- Create new configurations from report definitions
- Implement image report data provider as the initial consumer

This enhancement improves the flexibility of trace analysis by allowing users to generate and configure various report formats.

### How to test
Manual testing via trace-server:
1. Start the trace-server
2. Access the configuration interface
3. Create a new configuration using the Reports Data Provider:
   - Follow the standard configuration creation workflow
   - Select and configure the desired report type
   - Verify the configuration is saved correctly
   - Validate that an image report can be generated

Note: Unit tests will be added in upcoming commits.

### Follow-ups
Future work:
- Add comprehensive unit tests
- Implement providers for additional report types (text, HTML)
- Enhance validation for report configurations

### Review checklist

### DIFF
--- a/callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ImageReportDataProvider.java
+++ b/callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ImageReportDataProvider.java
@@ -1,0 +1,254 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Ericsson
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License 2.0 which
+ * accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.tracecompass.incubator.analysis.core.reports;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.nio.channels.FileChannel;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Path;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.jdt.annotation.NonNull;
+import org.eclipse.tracecompass.incubator.internal.analysis.core.Activator;
+import org.eclipse.tracecompass.tmf.core.config.ITmfConfigParamDescriptor;
+import org.eclipse.tracecompass.tmf.core.config.ITmfConfiguration;
+import org.eclipse.tracecompass.tmf.core.config.ITmfConfigurationSourceType;
+import org.eclipse.tracecompass.tmf.core.config.TmfConfigParamDescriptor;
+import org.eclipse.tracecompass.tmf.core.config.TmfConfiguration;
+import org.eclipse.tracecompass.tmf.core.config.TmfConfigurationSourceType;
+import org.eclipse.tracecompass.tmf.core.dataprovider.IDataProviderDescriptor;
+import org.eclipse.tracecompass.tmf.core.dataprovider.IDataProviderDescriptor.ProviderType;
+import org.eclipse.tracecompass.tmf.core.exceptions.TmfConfigurationException;
+import org.eclipse.tracecompass.tmf.core.model.DataProviderCapabilities;
+import org.eclipse.tracecompass.tmf.core.model.DataProviderDescriptor;
+import org.eclipse.tracecompass.tmf.core.trace.ITmfTrace;
+
+/**
+ * Handler for image-specific report configurations
+ *
+ * @author Kaveh Shahedi
+ */
+public class ImageReportDataProvider implements IReportDataProvider {
+
+    private static final String[] VALID_EXTENSIONS = { "png", "jpg", "jpeg", "svg" }; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
+                                                                                      // //$NON-NLS-5$
+    private static final String PATH = "path"; //$NON-NLS-1$
+    private static final String PATH_DESCRIPTION = "The path to the image file"; //$NON-NLS-1$
+
+    private static final ITmfConfigurationSourceType CONFIGURATION_SOURCE_TYPE;
+    private static final List<@NonNull ITmfConfigParamDescriptor> CONFIGURATION_DESCRIPTORS;
+    private static final String CONFIGURATION_ID = "org.eclipse.tracecompass.incubator.analysis.core.reports.image.config"; //$NON-NLS-1$
+    private static final String CONFIGURATION_NAME = "Image Report Configurator"; //$NON-NLS-1$
+    private static final String CONFIGURATION_DESCRIPTION = "Configure custom image reports (e.g., png, jpg, etc.)"; //$NON-NLS-1$
+
+    static {
+        CONFIGURATION_DESCRIPTORS = List.of(new TmfConfigParamDescriptor.Builder()
+                .setKeyName(PATH)
+                .setDescription(PATH_DESCRIPTION)
+                .setIsRequired(true)
+                .build());
+
+        CONFIGURATION_SOURCE_TYPE = new TmfConfigurationSourceType.Builder()
+                .setId(CONFIGURATION_ID)
+                .setName(CONFIGURATION_NAME)
+                .setDescription(CONFIGURATION_DESCRIPTION)
+                .setConfigParamDescriptors(Objects.requireNonNull(CONFIGURATION_DESCRIPTORS))
+                .build();
+    }
+
+    /**
+     * Create a descriptor for an image configuration
+     *
+     * @param configuration
+     *            Base configuration
+     * @return Image data provider descriptor
+     * @throws TmfConfigurationException
+     *             If configuration is invalid
+     */
+    @SuppressWarnings("null")
+    @Override
+    public @NonNull IDataProviderDescriptor createDescriptor(@NonNull ITmfTrace trace, @NonNull ITmfConfiguration configuration) throws TmfConfigurationException {
+        validateConfiguration(configuration);
+
+        // Create configuration with image-specific description if needed
+        String description = configuration.getDescription();
+        if (description.equals(TmfConfiguration.UNKNOWN)) {
+            description = configuration.getName();
+        }
+
+        IStatus status = createImage(trace, configuration);
+        if (!status.isOK()) {
+            throw new TmfConfigurationException(status.getMessage());
+        }
+
+        return getDescriptorFromConfig(trace, new TmfConfiguration.Builder()
+                .setId(configuration.getId())
+                .setSourceTypeId(configuration.getSourceTypeId())
+                .setName(configuration.getName())
+                .setDescription(description)
+                .setParameters(configuration.getParameters())
+                .build());
+    }
+
+    /**
+     * Create a descriptor from an existing configuration
+     *
+     * @param configuration
+     *            The configuration
+     * @return The descriptor
+     */
+    @Override
+    public @NonNull IDataProviderDescriptor getDescriptorFromConfig(@NonNull ITmfTrace trace, @NonNull ITmfConfiguration configuration) {
+        return new DataProviderDescriptor.Builder()
+                .setId(configuration.getId())
+                .setName(configuration.getName())
+                .setDescription(configuration.getDescription())
+                .setProviderType(ProviderType.NONE)
+                .setConfiguration(configuration)
+                .setCapabilities(new DataProviderCapabilities.Builder().setCanDelete(true).build())
+                .build();
+    }
+
+    /**
+     * Create a copy of the image file in the reports directory
+     *
+     * @param trace
+     *            The trace
+     * @param configuration
+     *            The configuration
+     * @return Status of the operation
+     */
+    @SuppressWarnings("null")
+    private static IStatus createImage(ITmfTrace trace, ITmfConfiguration configuration) {
+        IPath originalPath = new Path((String) configuration.getParameters().get(PATH));
+
+        if (!originalPath.toFile().exists()) {
+            return new Status(IStatus.ERROR, Activator.PLUGIN_ID, "The given image path does not exist"); //$NON-NLS-1$
+        }
+
+        if (originalPath.toFile().isDirectory()) {
+            return new Status(IStatus.ERROR, Activator.PLUGIN_ID, "The given image path is a directory"); //$NON-NLS-1$
+        }
+
+        String extension = originalPath.getFileExtension();
+        if (extension == null || !extension.matches(String.join("|", VALID_EXTENSIONS))) { //$NON-NLS-1$
+            return new Status(IStatus.ERROR, Activator.PLUGIN_ID, "Invalid image file extension. Supported extensions are: " //$NON-NLS-1$
+                    + String.join(", ", VALID_EXTENSIONS)); //$NON-NLS-1$
+        }
+
+        IPath reportsPath = ReportsDataProviderFactory.getConfigurationBasePath(trace, ReportsDataProviderFactory.getConfigurationParent(trace, configuration))
+                .addTrailingSeparator().append(configuration.getId());
+
+        File fromFile = originalPath.toFile();
+        File toFile = reportsPath.addTrailingSeparator().append(configuration.getName()).addFileExtension(extension).toFile();
+
+        try {
+            File parentDir = toFile.getParentFile();
+            if (!parentDir.exists()) {
+                boolean success = parentDir.mkdirs();
+                if (!success) {
+                    throw new IllegalStateException("Directory " + parentDir.getAbsolutePath() + " failed to create."); //$NON-NLS-1$ //$NON-NLS-2$
+                }
+            }
+
+            if (!toFile.exists()) {
+                boolean success = toFile.createNewFile();
+                if (!success) {
+                    throw new IllegalStateException("File " + toFile.getAbsolutePath() + " failed to create."); //$NON-NLS-1$ //$NON-NLS-2$
+                }
+            }
+        } catch (IOException e) {
+            String error = "Failed to create destination file"; //$NON-NLS-1$
+            return new Status(IStatus.ERROR, Activator.PLUGIN_ID, error, e);
+        }
+
+        try (FileInputStream fis = new FileInputStream(fromFile);
+                FileOutputStream fos = new FileOutputStream(toFile);
+                FileChannel source = fis.getChannel();
+                FileChannel destination = fos.getChannel()) {
+            destination.transferFrom(source, 0, source.size());
+        } catch (IOException e) {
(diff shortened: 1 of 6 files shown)


bhufmann (changes requested):
I did a first pass, mainly on the design level. I provided some feedback on the data provider creation in general and for containers/images DPs. Please have a look. If you have any question please let me know.

callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ReportsDataProviderFactory.java
```diff
@@ -0,0 +1,729 @@
...
+
+            result.add(new TmfConfigurationSourceType.Builder()
+                    .setName(sourceType.getName())
+                    .setDescription(sourceType.getDescription())
+                    .setId(sourceType.getId())
+                    .setConfigParamDescriptors(allDescriptors)
+                    .build());
+        }
+
+        return result;
```
bhufmann, line 321: listing the source types I see a duplicate entry:
```
[
    {
        "id": "org.eclipse.tracecompass.incubator.analysis.core.reports.config",
        "name": "Reports Configurator",
        "description": "Configure custom reports",
        "parameterDescriptors": [
            {
                "keyName": "reportType",
                "description": "Should be NONE",
                "dataType": "STRING",
                "isRequired": true
            },
            {
                "keyName": "parentId",
                "description": "ID of the parent report",
                "dataType": "STRING",
                "isRequired": true
            }
        ],
        "schema": "None"
    },
    {
        "id": "org.eclipse.tracecompass.incubator.analysis.core.reports.config",
        "name": "Reports Configurator",
        "description": "Configure custom reports",
        "parameterDescriptors": [
            {
                "keyName": "reportType",
                "description": "Should be NONE",
                "dataType": "STRING",
                "isRequired": true
            },
            {
                "keyName": "parentId",
                "description": "ID of the parent report",
                "dataType": "STRING",
                "isRequired": true
            }
        ],
        "schema": "None"
    },
]
```

callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ReportsDataProviderFactory.java
```diff
@@ -0,0 +1,729 @@
...
+            ITmfConfigurationSourceType sourceType = provider.getConfigurationSourceType();
+            List<ITmfConfigParamDescriptor> allDescriptors = new ArrayList<>(sourceType.getConfigParamDescriptors());
+            allDescriptors.addAll(List.of(
+                    new TmfConfigParamDescriptor.Builder()
+                            .setKeyName(REPORT_TYPE)
+                            .setDescription(NLS.bind(REPORT_TYPE_DESC, provider.getReportType().name().toUpperCase()))
+                            .setIsRequired(true)
+                            .build(),
+                    new TmfConfigParamDescriptor.Builder()
+                            .setKeyName(PARENT_ID)
```
bhufmann, line 308: So, the user has to provide the parentID, to group it under, right? It's the id that the user used to create the container (see above). The description should be clear about it.

callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ReportsDataProviderFactory.java
```diff
@@ -0,0 +1,729 @@
...
+                .setConfigParamDescriptors(baseDescriptors)
+                .build());
+
+        // Add provider configuration types with the same descriptor structure
+        for (IReportDataProvider provider : ReportsDataProviderRegistry.getProviders()) {
+            ITmfConfigurationSourceType sourceType = provider.getConfigurationSourceType();
+            List<ITmfConfigParamDescriptor> allDescriptors = new ArrayList<>(sourceType.getConfigParamDescriptors());
+            allDescriptors.addAll(List.of(
+                    new TmfConfigParamDescriptor.Builder()
+                            .setKeyName(REPORT_TYPE)
```
bhufmann, line 303: The report type is not needed, because the configuration source type is connected this provider type. Or do I misunderstand?

callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ReportsDataProviderFactory.java
```diff
@@ -0,0 +1,729 @@
...
+    @SuppressWarnings("null")
+    @Override
+    public @NonNull List<@NonNull ITmfConfigurationSourceType> getConfigurationSourceTypes() {
+        List<ITmfConfigurationSourceType> result = new ArrayList<>();
+
+        // Add base configuration source type with the common descriptors
+        List<ITmfConfigParamDescriptor> baseDescriptors = new ArrayList<>(CONFIGURATION_SOURCE_TYPE.getConfigParamDescriptors());
+        baseDescriptors.addAll(List.of(
+                new TmfConfigParamDescriptor.Builder()
+                        .setKeyName(REPORT_TYPE)
```
bhufmann, line 280: The report type is not needed, because the configuration source type is connected this provider type. Or do I misunderstand?

callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ReportsDataProviderFactory.java
bhufmann: Also, what about making it optional? If omitted then the report is put directly under the reportsDatatProviderFactory.

callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ReportsDataProviderFactory.java
```diff
@@ -0,0 +1,729 @@
...
+                        .setDescription(NLS.bind(REPORT_TYPE_DESC, ProviderType.NONE.name().toUpperCase()))
+                        .setIsRequired(true)
+                        .build(),
+                new TmfConfigParamDescriptor.Builder()
+                        .setKeyName(PARENT_ID)
+                        .setDescription(PARENT_ID_DESC)
```
bhufmann, line 286: So, the user has to provide a unique parentID, right? This parentID is the ID of the container created. You should mention in the description that it needs to be unique.

callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ReportsDataProviderFactory.java
```diff
@@ -0,0 +1,729 @@
...
+
+    @Override
+    public ITmfConfigurationSourceType getConfigurationSourceType() {
+        return CONFIGURATION_SOURCE_TYPE;
+    }
+
+    @Override
+    public @NonNull IDataProviderDescriptor createDescriptor(@NonNull ITmfTrace trace, @NonNull ITmfConfiguration configuration) throws TmfConfigurationException {
+        // Validate the configuration
+        ITmfConfiguration config = validateConfiguration(trace, configuration);
```
bhufmann, line 337: I'm trying to create container with the following input parameters:

```
{
   "name": "Main Container",
   "description" : "The main container",
   "sourceTypeId": "org.eclipse.tracecompass.incubator.analysis.core.reports.config",
   "parameters": {
       "parentId": "my.parent.id1",
       "reportType": "NONE" 
   }
}
```
but I get the following error:
`failed to create derived output: 400 Invalid parent ID`

It looks like that I can't provide my own parent ID. I understood that the idea of the parentID field was to provide an user-defined parentID, that can be passed to the image creation. 

If this is not the intention and we have to put the reportsDataProvider ID as parent then this is not practical for UI users because the ID is not visible to end-users. In that case it would be redundant also, because it can be assigned by the backend and doesn't need to be provided by the user.

So, I think if you'd like to continue with allowing to group reports, then the parentID needs to be an input for the container and then the user can use it when creating image reports. As I stated above, it could optional and if omitted the parent would be that root report data provider.

bhufmann (changes requested):

callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ReportsDataProviderFactory.java
kavehshahedi: Making it optional sounds pretty good. I'll investigate it.

callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ReportsDataProviderFactory.java
kavehshahedi: The `my.parent.id1` that you put in the parameter, does it come from a created data provider by reports factory? If yes, then the error that you are getting is weird and should be investigated. If no, you should put a proper and valid data provider ID (a sub-factory), which was created by the reports factory. 
So the steps would be:
1. If it's not set, the parentId will be the reports factory ID (root) -> Not in this commit yet, but I'll fix it.
2. Once a sub-factory report (i.e., container) is created, you capture its ID
3. If a new report (container or image or etc.) is created, we can use the ID from step 2 to put it as the parentId.

callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ReportsDataProviderFactory.java
```diff
@@ -0,0 +1,723 @@
...
+        if (configuration.getName().equals(TmfConfiguration.UNKNOWN) || configuration.getSourceTypeId().equals(TmfConfiguration.UNKNOWN)) {
+            throw new TmfConfigurationException("Missing configuration name or type");//$NON-NLS-1$
+        }
+
+        String parentId = (String) configuration.getParameters().get(PARENT_ID);
+        if (parentId == null) {
+            parentId = ID;
+        }
+
+        if (!parentId.equals(ID) && !fTmfConfigurationTable.contains(parentId, trace.getUUID())) {
```
bhufmann, line 443: This condition still fails when creating a container and providing a parentId and the TmfConfigurationException is created:

```
{ 
  "name": "Main Container", 
  "description" : "The main container",
   "sourceTypeId": "org.eclipse.tracecompass.incubator.analysis.core.reports.config", 
   "parameters": { 
     "parentId": "my.parent.id1", 
     "reportType": "NONE" 
   }
 }
```

`fTmfConfigurationTable` won't contain that parentID because it hasn't been submitted.

kavehshahedi: I didn't quite understand your issue here. The `my.parent.id1` should be a valid created configuration by the reports factory, and in case if not match these criteria, it will throw an exception. 

However, I realized that probably your issue is that the error message is not clear. So, I changed it from *Invalid parent ID* to *Parent configuration does not exist*.

bhufmann: My issue is that there is an exception. There should not be an exception. I'm creating a new parent and it should not be rejected with `Parent doesn't exist` and, instead, the new parent should be created.

kavehshahedi: No, the parentId parameter is the ID of the configuration that the configuration you are creating would be its parent. It doesn't mean that the configuration that you are creating will have this specific ID. 

In the example you mentioned, you are creating a report with the name "Main Container" and a specific description. Then, when you indicate the `parentId`, the factory looks for an already-created configuration (container, btw) with that specific ID, and if it exists, it will be assigned as its parent (root->parentId->new config). If it doesn't exist (you haven't created that parent configuration yet), then it will throw that error.

bhufmann: My idea of allowing the user to input the parentId during creation of a container which then the user can use to input as parentID when creating the report. The user would then know about the ID to use and would not have to look for it. If omitted during creation of the report then report would be directly under the report data provider factory. 

That wouldn't allow the deep hierarchy though out of the box.

bhufmann: Your comment https://github.com/eclipse-tracecompass-incubator/org.eclipse.tracecompass.incubator/pull/147#discussion_r1955200197 helped me going forward with testing and review. We can keep your proposal and provide the deep hierarchy.

callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ImageReportDataProvider.java
```diff
@@ -0,0 +1,254 @@
...
+
+    private static final String[] VALID_EXTENSIONS = { "png", "jpg", "jpeg", "svg" }; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
+                                                                                      // //$NON-NLS-5$
+    private static final String PATH = "path"; //$NON-NLS-1$
+    private static final String PATH_DESCRIPTION = "The path to the image file"; //$NON-NLS-1$
+
+    private static final ITmfConfigurationSourceType CONFIGURATION_SOURCE_TYPE;
+    private static final List<@NonNull ITmfConfigParamDescriptor> CONFIGURATION_DESCRIPTORS;
+    private static final String CONFIGURATION_ID = "org.eclipse.tracecompass.incubator.analysis.core.reports.image.config"; //$NON-NLS-1$
+    private static final String CONFIGURATION_NAME = "Image Configurator"; //$NON-NLS-1$
```
bhufmann, line 56: I wonder if it should have a different text. `Image Configurator` might not be clear. Maybe `Image Report Configurator`?

callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ReportsDataProviderFactory.java
```diff
@@ -0,0 +1,723 @@
...
+ *
+ * @author Kaveh Shahedi
+ */
+public class ReportsDataProviderFactory implements IDataProviderFactory, ITmfDataProviderConfigurator, IReportDataProvider {
+
+    /** Reports Provider factory ID */
+    public static final String ID = "org.eclipse.tracecompass.incubator.analysis.core.reports.reportsDataProviderFactory"; //$NON-NLS-1$
+
+    private static final String NAME = "Reports"; //$NON-NLS-1$
+    private static final String DESCRIPTION = "Configure various types of reports (text, image, etc.)"; //$NON-NLS-1$
```
bhufmann, line 77: Could you please also mention the possibility to create report containers? Otherwise it won't be known to the user. Maybe: `Configure various types of reports (text, image, etc.) or report folders for grouping purposes`. You could replace `folders` with `containers` or `groups` if you think it's clearer.

callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ReportsDataProviderFactory.java
bhufmann: My point is that the `reportType` (see below) can have only one value for the image type, which is `IMAGE`. In the description you event mention that it should be `IMAGE`. Besides there is a 1:1 relationship between the source type with id `org.eclipse.tracecompass.incubator.analysis.core.reports.image.config` to the `reportType` `IMAGE`. So, I don't see that users have to add the `reportType` because it's implied, no?

Note that the same comment is for the container and `reportType`  `NONE`.

```
    {
        "id": "org.eclipse.tracecompass.incubator.analysis.core.reports.image.config",
        "name": "Image Configurator",
        "description": "Configure custom image reports (e.g., png, jpg, etc.)",
        "parameterDescriptors": [
            {
                "keyName": "path",
                "description": "The path to the image file",
                "dataType": "STRING",
                "isRequired": true
            },
            {
                "keyName": "reportType",
                "description": "Should be IMAGE",
                "dataType": "STRING",
                "isRequired": true
            },
            {
                "keyName": "parentId",
                "description": "Unique ID of the parent report. The parent must be an existing report. If not specified, the root report container will be assigned as the parent.",
                "dataType": "STRING",
                "isRequired": false
            }
        ],
        "schema": "None"
    },

```

callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ReportsDataProviderFactory.java
```diff
@@ -0,0 +1,723 @@
...
+    private static final ITmfConfigurationSourceType CONFIGURATION_SOURCE_TYPE;
+    private static final String CONFIGURATION_ID = "org.eclipse.tracecompass.incubator.analysis.core.reports.config"; //$NON-NLS-1$
+    private static final String CONFIGURATION_NAME = "Reports Configurator"; //$NON-NLS-1$
+    private static final String CONFIGURATION_DESCRIPTION = "Configure custom reports"; //$NON-NLS-1$
+
+    /** Constants for configuration parameters */
+    private static final String REPORT_TYPE = "reportType"; //$NON-NLS-1$
+    private static final String REPORT_TYPE_DESC = "Should be {0}"; //$NON-NLS-1$
+    private static final String PARENT_ID = "parentId"; //$NON-NLS-1$
+    private static final String PARENT_ID_DESC = "Unique ID of the parent report. The parent must be an existing report. If not specified, the root report container will be assigned as the parent."; //$NON-NLS-1$
```
bhufmann, line 88: Sorry, my previous comment about mentioning unique ID was because I understood it would be user provider (i.e. user defined ID). As we now established, the ID is generated by the back-end, then the backend will take care of the uniqueness. So, you could make the description shorter again.

bhufmann (changes requested):
I was able to test the creation of containers, sub-containers and images, their deletion and some error case with `tsp-python-client` and it works well. The code is clean as well. I have a couple of questions/comments to follow-up.

bhufmann (approved):
Thank you very much for this contribution. It's working well and is a good example on how the customization APIs can be used. Looks good to me.

MatthewKhouzam (approved):

PR #151, opened 2025-02-18
Use trace-event-logger lib

PatrickTasse:
### What it does

Replace all use of TraceCompassLogUtils with LogUtils from the trace-event-logger library.

Update all targets to include the trace-event-logger jar from Maven.

Add trace-event-logger plug-in to RCP feature and releng-site.

Add trace-event-logger plug-in to trace-server product.

### How to test

Enable logging.

### Follow-ups

N/A

### Review checklist


bhufmann (changes requested):
the trace-event-logger needs to be added to the traceserver.product

bhufmann (approved):
Thanks for the update. Looks good to me.

PR #163, opened 2025-03-11
inandout: do not share configuration table

frallax:
### What it does

Make fTmfConfigurationTable simply private.
This avoids that multiple instances of the same factory class can share the table.

Unit test updated accordingly.

### How to test

Make fTmfConfigurationTable simply private, and then run the test as is on master. It will fail since a different instance of InAndOutDataProviderFactory is returned.

### Follow-ups

None

### Review checklist


bhufmann (approved):
Thanks for the fix. Thanks!

MatthewKhouzam (approved):

PR #166, opened 2025-03-12
inandout: Use config separator in analysis id and output id

PatrickTasse:
### What it does

When a configuration is applied, use the CONFIG_SEPARATOR '+' before the configId in the analysisId and outputId.

### How to test

Create derived output from the factory and examine the analysisId and outputId.

### Follow-ups

N/A

### Review checklist


bhufmann (approved):
Makes sense to have a specific separator for configurations. Not convinced that we need to have an API in TC mainline for it, but I'm ok to do so.

PR #174, opened 2025-03-27
server: Add get image data provider endpoint

kavehshahedi:
### What it does

Following the introduction of configurable reports data providers, the trace server can now face a new data provider type for image providers (i.e., "IMAGE"). This commit introduces the required endpoint for fetching the images from the trace server based on the given configurable output id.

### How to test

In order to fetch a specific image output from the server, you can send a `GET` request with the following structure:

`/tsp/api/experiments/EXP-ID/outputs/image/IMAGE-OUTPUT-ID`

### Follow-ups
1. Proper unit tests should be implemented
2. Front-end products should follow this endpoint for fetching the images

### Review checklist

### DIFF
--- a/callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ImageReportDataProvider.java
+++ b/callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ImageReportDataProvider.java
@@ -118,7 +118,7 @@ public class ImageReportDataProvider implements IReportDataProvider {
                 .setId(configuration.getId())
                 .setName(configuration.getName())
                 .setDescription(configuration.getDescription())
-                .setProviderType(ProviderType.NONE)
+                .setProviderType(ProviderType.IMAGE)
                 .setConfiguration(configuration)
                 .setCapabilities(new DataProviderCapabilities.Builder().setCanDelete(true).build())
                 .build();
--- a/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
+++ b/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
@@ -35,6 +35,7 @@ import static org.eclipse.tracecompass.incubator.internal.trace.server.jersey.re
 import static org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core.services.EndpointConstants.EXP_UUID;
 import static org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core.services.EndpointConstants.FILTER_QUERY_PARAMETERS;
 import static org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core.services.EndpointConstants.FILTER_QUERY_PARAMETERS_EX;
+import static org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core.services.EndpointConstants.IMG;
 import static org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core.services.EndpointConstants.INDEX;
 import static org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core.services.EndpointConstants.INDEX_EX;
 import static org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core.services.EndpointConstants.INVALID_PARAMETERS;
@@ -74,6 +75,9 @@ import static org.eclipse.tracecompass.incubator.internal.trace.server.jersey.re
 import static org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core.services.EndpointConstants.VTB;
 import static org.eclipse.tracecompass.incubator.internal.trace.server.jersey.rest.core.services.EndpointConstants.X_Y;
 
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -1098,6 +1102,70 @@ public class DataProviderService {
             TmfModelResponse<?> treeResponse = provider.fetchTree(params, null);
             Object model = treeResponse.getModel();
             return Response.ok(model instanceof TmfTreeModel ? new TmfModelResponse<>(new TreeModelWrapper((TmfTreeModel<@NonNull ITmfTreeDataModel>) model), treeResponse.getStatus(), treeResponse.getStatusMessage()) : treeResponse).build();
+        }
+    }
+
+    /**
+     * Query the provider for an image-based data provider
+     *
+     * @param expUUID
+     *            desired experiment UUID
+     * @param outputId
+     *            Output ID for the data provider to query
+     * @return {@link Response} with the corresponding image
+     */
+    @GET
+    @Path("/image/{outputId}")
+    @Tag(name = IMG)
+    @Produces({ MediaType.APPLICATION_OCTET_STREAM, MediaType.APPLICATION_JSON })
+    @Operation(summary = "API to get an image", responses = {
+            @ApiResponse(responseCode = "200", description = "Returns the image file", content = @Content(mediaType = "application/octet-stream")),
+            @ApiResponse(responseCode = "400", description = MISSING_PARAMETERS, content = @Content(schema = @Schema(implementation = String.class))),
+            @ApiResponse(responseCode = "404", description = PROVIDER_NOT_FOUND, content = @Content(schema = @Schema(implementation = String.class))),
+            @ApiResponse(responseCode = "500", description = "Error reading the image file", content = @Content(schema = @Schema(implementation = String.class)))
+    })
+    public Response getImage(
+            @Parameter(description = EXP_UUID) @PathParam("expUUID") UUID expUUID,
+            @Parameter(description = OUTPUT_ID) @PathParam("outputId") String outputId) {
+
+        if (outputId == null) {
+            return Response.status(Status.BAD_REQUEST).entity(MISSING_OUTPUTID).build();
+        }
+
+        TmfExperiment experiment = ExperimentManagerService.getExperimentByUUID(expUUID);
+        if (experiment == null) {
+            return Response.status(Status.NOT_FOUND).entity(NO_SUCH_TRACE).build();
+        }
+
+        IDataProviderDescriptor descriptor = getDescriptor(experiment, outputId);
+        if (descriptor == null || descriptor.getType() != ProviderType.IMAGE) {
+            return Response.status(Status.NOT_FOUND).entity(NO_SUCH_PROVIDER).build();
+        }
+
+        ITmfConfiguration config = descriptor.getConfiguration();
+        if (config == null) {
+            return Response.status(Status.NOT_FOUND).entity("No configuration found for this provider").build(); //$NON-NLS-1$
+        }
+
+        String imagePath = (String) config.getParameters().get("path"); //$NON-NLS-1$
+        if (imagePath == null) {
+            return Response.status(Status.NOT_FOUND).entity("Image path not found in configuration").build(); //$NON-NLS-1$
+        }
+
+        File imageFile = new File(imagePath);
+        if (!imageFile.exists() || !imageFile.isFile()) {
+            return Response.status(Status.NOT_FOUND).entity("Image file not found").build(); //$NON-NLS-1$
+        }
+
+        try {
+            String contentType = Files.probeContentType(imageFile.toPath());
+            if (contentType == null) {
+                contentType = MediaType.APPLICATION_OCTET_STREAM;
+            }
+
+            return Response.ok(imageFile, contentType).build();
+        } catch (IOException e) {
+            return Response.status(Status.BAD_REQUEST).entity(e.getMessage()).build();
         }
     }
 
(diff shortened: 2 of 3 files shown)


trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
```diff
@@ -1101,6 +1105,70 @@ private Response getTree(UUID expUUID, String outputId, QueryParameters queryPar
...
+     * Query the provider for an image-based data provider
+     *
+     * @param expUUID
+     *            desired experiment UUID
+     * @param outputId
+     *            Output ID for the data provider to query
+     * @return {@link Response} with the corresponding image
+     */
+    @GET
+    @Path("/image/{outputId}")
```
bhufmann, line 1118: I'm debating with myself it this is the right endpoint. `/image/` for images per experiment. We would we have then other endpoints for other types, e.g. `/pdf/` or `/html` (multipart) or others.  See also comment on 
https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/242#issuecomment-2783260628

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
```diff
@@ -1101,6 +1105,83 @@ private Response getTree(UUID expUUID, String outputId, QueryParameters queryPar
...
+     * @param expUUID
+     *            desired experiment UUID
+     * @param outputId
+     *            Output ID for the data provider to query
+     * @return {@link Response} with the corresponding report data
+     */
+    @GET
+    @Path("/report/{outputId}")
+    @Tag(name = RPT)
+    @Produces({ MediaType.APPLICATION_OCTET_STREAM, MediaType.APPLICATION_JSON })
```
bhufmann, line 1122: Do we need MediaType.APPLICATION_JSON? In what cases will it return a JSON reply?

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
```diff
@@ -1101,6 +1105,83 @@ private Response getTree(UUID expUUID, String outputId, QueryParameters queryPar
...
+    @Operation(summary = "API to get an image", responses = {
+            @ApiResponse(responseCode = "200", description = "Returns the image file", content = @Content(mediaType = "application/octet-stream")),
```
bhufmann, line 1124: This implies that it's always an image. It should be more generic

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
```diff
@@ -1101,6 +1105,83 @@ private Response getTree(UUID expUUID, String outputId, QueryParameters queryPar
...
+            @ApiResponse(responseCode = "400", description = MISSING_PARAMETERS, content = @Content(schema = @Schema(implementation = String.class))),
+            @ApiResponse(responseCode = "404", description = PROVIDER_NOT_FOUND, content = @Content(schema = @Schema(implementation = String.class))),
```
bhufmann, line 1126: There needs to be a different description string because the error also could be for the unknown sub-type.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
```diff
@@ -1101,6 +1105,83 @@ private Response getTree(UUID expUUID, String outputId, QueryParameters queryPar
...
+        if (descriptor == null || descriptor.getType() != ProviderType.REPORT) {
+            return Response.status(Status.NOT_FOUND).entity(NO_SUCH_PROVIDER).build();
+        }
+
+        ProviderType reportType = descriptor.getSubType();
+        if (reportType == null) {
+            return Response.status(Status.NOT_FOUND).entity(NO_SUCH_PROVIDER).build();
+        }
+
+        ITmfConfiguration config = descriptor.getConfiguration();
```
bhufmann, line 1152: We have an issue here. The report is a customization for a that is specific for the Trace Compass server and won't exists in other server implementation. This endpoint needs to be independent from it. 

This endpoint queries a specific data provider and this data provider has to know what it supposed to return in this endpoint, so the path should come from the data provider.

MatthewKhouzam (approved):
LGTM

image


You need to add the depends on pull request from mainline.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
```diff
@@ -1101,6 +1105,87 @@ private Response getTree(UUID expUUID, String outputId, QueryParameters queryPar
...
+     * @param expUUID
+     *            desired experiment UUID
+     * @param outputId
+     *            Output ID for the data provider to query
+     * @return {@link Response} with the corresponding report data
+     */
+    @GET
+    @Path("/report/{outputId}")
+    @Tag(name = RPT)
+    @Produces({MediaType.APPLICATION_OCTET_STREAM, MediaType.TEXT_HTML, MediaType.TEXT_PLAIN })
```
bhufmann, line 1122: In you report related code, you call `Files.probeContentType` which will return `image/jpg` (for example). Will that work?

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
```diff
@@ -1101,6 +1105,87 @@ private Response getTree(UUID expUUID, String outputId, QueryParameters queryPar
...
+            if (reportResponse.getStatus() != ITmfResponse.Status.COMPLETED) {
+                return Response.status(Status.INTERNAL_SERVER_ERROR).entity(reportResponse.getStatusMessage()).build();
+            }
+
+            TmfDataProviderConfigurationDataModel responseModel = reportResponse.getModel();
+            if (responseModel == null) {
+                return Response.status(Status.NOT_FOUND).entity("Report data model not found").build(); //$NON-NLS-1$
+            }
+
+            return Response.ok(responseModel.getContent(), responseModel.getContentType()).build();
```
bhufmann, line 1181: For images, responseModel.getContent() returns a object of type `File`.  The content type will be determined by `Files.probeContentType()` which could return `image/jpg` or `image/svg` or if it can't be determined it will be `application/octet-stream`. I'm clear what will happen in each case and how a FE client can handle that.

kavehshahedi: Now the endpoint is producing `MediaType.WILD_CARD` to ensure any MIME data can be handled. The `responseModel.getContentType` would be responsible for indicating the output type (e.g., `image/jpg`).
The FE client should handle the output based on the response's headers that indicate the output type.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
```diff
@@ -1101,6 +1105,87 @@ private Response getTree(UUID expUUID, String outputId, QueryParameters queryPar
...
+        if (descriptor.getType() != ProviderType.MIME) {
+            return Response.status(Status.NOT_FOUND).entity("The requested output is not a MIME report").build(); //$NON-NLS-1$"
+        }
+
+        ITmfConfiguration config = descriptor.getConfiguration();
+        if (config == null) {
+            return Response.status(Status.NOT_FOUND).entity(NO_SUCH_CONFIGURATION).build();
+        }
+
+        IDataProviderFactory factory = manager.getFactory(REPORTS_FACTORY_ID);
```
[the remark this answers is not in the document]
kavehshahedi: Here, I tried to decouple this endpoint from `REPORTS_FACTORY_ID` by using the given output's id in the endpoint. However, the issue is that, except for `ReportsDataProviderFactory`, no other report data provider classes are inheriting the `IDataProviderFactory`. Hence, when loading the available factories in the startup, the report-based data providers whose ID is not `REPORTS_FACTORY_ID` would not be recognized as a factory. 

As the `ReportsDataProviderFactory` class is responsible for handling the entire pipeline of reports, I guess we might need to keep the structure as is. Also, the endpoint is specifically for getting report data, which makes it suitable for using the `REPORTS_FACTORY_ID` in it, although in some cases, the user might not have installed this plugin. 

What do you think of this?

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
bhufmann: How, will it work? Clients could ask for e.g. applicationt/pdf and the endpoint will allow it because of the wildcard but then the implementation won't be able to create a PDF but an image instead. I'm not clear about the procedure and header negotiation yet.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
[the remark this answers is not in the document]
kavehshahedi: I guess that's fine. Since the endpoint is only for this specific DP, in case of missing (e.g., not being installed), we can tell the user that it's not installed as a plug-in. wdyt?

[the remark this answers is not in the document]
bhufmann: > I guess that's fine. Since the endpoint is only for this specific DP, in case of missing (e.g., not being installed), we can tell the user that it's not installed as a plug-in. wdyt?

That's not really an option. Users cannot install plug-ins. Trace server applications are pre-build.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
kavehshahedi: So, basically, the client can ask only based on the report type. For instance, the client only knows that it's an image, but doesn't know if it's a JPG or PNG or SVG. The endpoint then fetches the DP data, and returns the output with a specific content type, which can be image/jpg or image/png or etc. This information is provided in the response's headers as the content type (I have checked it already).
So, it would be the responsibility of the FE to properly show the image in the output based on the response's content type.

callstack/org.eclipse.tracecompass.incubator.analysis.core/src/org/eclipse/tracecompass/incubator/analysis/core/reports/ImageReportDataProvider.java
```diff
@@ -33,17 +36,22 @@
...
+import org.eclipse.tracecompass.tmf.core.response.TmfModelResponse;
 import org.eclipse.tracecompass.tmf.core.trace.ITmfTrace;
 
 /**
  * Handler for image-specific report configurations
  *
  * @author Kaveh Shahedi
  */
-public class ImageReportDataProvider implements IReportDataProvider {
+public class ImageReportDataProvider implements IReportDataProvider, IDataProviderFactory {
```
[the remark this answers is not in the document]
kavehshahedi: You are absolutely right. I was working around with this `IDataProiderFactory` to see how we can solve the `manager.getFactory()` issue in the `DataProviderService` class, but I forgot to remove it in the end from the class. I will fix it.

bhufmann:
Please pay attention to PR #188 where the o.e.tc.incubator.analysis.* plug-ins are move to the analysis folder, in order to remove the callstack folder after cleaning-up that folder (#185). You will have re-import the plug-ins to your development environment and apply your report related changes there..

PR #176, opened 2025-04-01
server: Validate parent descriptor when creating derived data provider

PatrickTasse:
### What it does

Validate parent descriptor when creating derived data provider.

This can prevent creating derived providers from hidden data providers.

### How to test

With https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/243, configure the trace-server to hide a data provider configurator.

For example in ~/.tracecompass-webapp/.metadata/.plugins/org.eclipse.tracecompass.tmf.core/dataprovider.ini, add line:

hide:org.eclipse.tracecompass.incubator.inandout.core.analysis.inAndOutdataProviderFactory:*

Then try to create a derived data provider. For example using tsp-python-client:

./tsp_cli_client --create-output org.eclipse.tracecompass.incubator.inandout.core.analysis.inAndOutdataProviderFactory --uuid _uuid_ --json-file _jsonfile_

It should fail with 404 error code (Data provider doesn't exist).

### Follow-ups

N/A

### Review checklist

### DIFF
--- a/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
+++ b/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
@@ -1302,6 +1302,11 @@ public class DataProviderService {
                 return errorResponse;
             }
 
+            IDataProviderDescriptor parentDescriptor = getDescriptor(experiment, outputId);
+            if (parentDescriptor == null) {
+                return Response.status(Status.NOT_FOUND).entity(NO_SUCH_PROVIDER + ": " + outputId).build(); //$NON-NLS-1$
+            }
+
             IDataProviderFactory factory = manager.getFactory(outputId);
             if (factory == null) {
                 return Response.status(Status.NOT_FOUND).entity(NO_SUCH_PROVIDER).build();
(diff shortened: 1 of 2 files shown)


trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
```diff
@@ -1302,6 +1302,11 @@ public Response createProvider(
                 return errorResponse;
             }
 
+            IDataProviderDescriptor parentDescriptor = getDescriptor(experiment, outputId);
```
bhufmann, line 1305: line 1310 has the same effect. I don't think it's needed. You could update the error message below to add the output ID to it the error.

bhufmann (changes requested):

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
PatrickTasse: In an earlier version of the TC PR, the factories weren't being filtered-out at all so this was necessary to prevent creation of derived provider from a hidden configurator.

In the latest TC PR, factories can be filtered-out and would prevent the creation, but it only works when the hiding configuration applies to all trace types.

So I believe that this is still necessary when the hiding configuration is only for a specific trace type.

bhufmann: It's fine like that. It's inline for other custom data provider endpoints.

PR #179, opened 2025-04-11
Remove customized data providers from manager

### LINKED ISSUE #178: Custom InAndOutDataProvider not removed from DataProviderManager
### Bug Description:

When creating InAndOut configuration from the tsp-python-client it will create derived data providers (e.g. Call Stack View). After opening the data provider once, delete customization, re-create the same one and open the call stack view again will cause an empty view. The reason is that the dp instance was not removed after deletion of the customization

### Steps to Reproduce:

1. Create custom InAndOut using tsp-python-client
2. Open InAndOut call stack view from vscode-trace-extension
3. Delete custom InAndOut using tsp-python-client
4. Re-create same custom InAndOut as in 1)
5. Open InAndOut call stack view from vscode-trace-extension
6. The view is empty

### Additional Information

- Operating System: Any supported OS
- Trace Compass Version: 10.3

bhufmann:
### What it does

Remove any instances of derived data providers and sub-data providers created with InAndAnalysis. 

fixes #178 

Requires the following Trace Compass PRs:
- https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/254
- https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/256
- https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/257

### How to test
1. Create custom InAndOut using tsp-python-client
2. Open InAndOut call stack view from vscode-trace-extension
3. Delete custom InAndOut using tsp-python-client
4. Re-create same custom InAndOut as in 1)
5. Open InAndOut call stack view from vscode-trace-extension
6. The view is not empty 

Note: The same issue exists for the 3 other derived views and it is fixed with this PR.

### Review checklist

### DIFF
--- a/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
+++ b/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
@@ -1318,7 +1318,16 @@ public class DataProviderService {
             if (configurator == null) {
                 return Response.status(Status.NOT_FOUND).entity(NO_SUCH_PROVIDER).build();
             }
+
+            // Get all descriptors of all corresponding derived data providers
+            List<IDataProviderDescriptor> allDerivedDps = manager.getAvailableProviders(experiment, derivedDescriptor.getConfiguration());
+
+            // Remove any derived data provider instances stored in the manager
+            allDerivedDps.forEach(desc -> manager.removeDataProvider(experiment, desc.getId()));
+
+            // Clean-up configuration
             configurator.removeDataProviderDescriptor(experiment, derivedDescriptor);
+
             return Response.ok(derivedDescriptor).build();
         } catch (TmfConfigurationException e) {
             return Response.status(Status.BAD_REQUEST).entity(e.getMessage()).build();


analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutDataProviderFactory.java
```diff
@@ -245,17 +253,61 @@ private void applyConfiguration(ITmfTrace trace, ITmfConfiguration config, boole
...
+            fDataProviderPrefixes = DataProviderManager.getInstance().getAvailableProviders(trace)
+                    .stream()
+                    .filter(predicate)
+                    .map(mapper)
+                    .collect(Collectors.toList());
+        }
+    }
+
+    private void removeChildrenDataProviders(ITmfTrace trace, ITmfConfiguration config) {
+        List<String> dpPrefixes = fDataProviderPrefixes;
```
PatrickTasse, line 304: I wonder, instead of computing the list of prefixes at creation, wouldn't it be simpler at removal to go through all available providers and remove any found whose id ends with +<configId> ?

bhufmann: I had that initially. I changed it because I have to call getAvailableDataProviders() and find all relevant data providers for each trace and the experiment. I felt that was inefficient. WDYT?

PatrickTasse: It seems the inefficiency at removal is the same as what would be avoided at creation, and we avoid having to extract and store the prefixes?

bhufmann: The storing of prefixes is done only once.

bhufmann: I'll change it back. I think added complexity is not really warrented.

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutDataProviderFactory.java
```diff
@@ -248,16 +251,37 @@ private void applyConfiguration(ITmfTrace trace, ITmfConfiguration config, boole
...
             }
-            // only remove for traces in experiment
+            // Only remove configuration for traces in experiment
             return;
         }
+        // This will remove the analysis module from the trace and disposes it
         InAndOutAnalysisModule.remove(config, trace);
     }
 
+    private static void removeChildrenDataProviders(ITmfTrace trace, ITmfConfiguration config) {
```
PatrickTasse, line 268: Yes, I like this implementation a lot better. But I can't help but think that this would need to be done by all factories that support configurations. It could have been the new API in DataProviderManager: removeProviders(ITmfTrace, ITmfConfiguration), that would remove all data providers that have this configuration, instead of per data provider id.

I also wonder if, in this implementation, the first call to the method would also cover the removal of the data provider that is being explicitly removed by the DataProviderService in the previous commit.

So I'm thinking, in DataProviderService.deleteDerivedProvider(), could it just call something like DataProviderManager.removeProviders(ITmfTrace, ITmfConfiguration), and then the configurators/factories only need to worry about removing the descriptor and configuration? They are not really the owners of the data provider instances.

One consideration is that either DataProviderService or DataProviderManager would need to do this processing for the full trace set (experiment and its child traces).

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutDataProviderFactory.java
```diff
@@ -248,16 +251,37 @@ private void applyConfiguration(ITmfTrace trace, ITmfConfiguration config, boole
...
+        Predicate<IDataProviderDescriptor> predicate = desc -> {
+            ITmfConfiguration cfg = desc.getConfiguration();
+            return (cfg != null &&
+                    cfg.getSourceTypeId().equals(SegmentSpecifierConfiguration.IN_AND_OUT_CONFIG_SOURCE_TYPE_ID) &&
```
PatrickTasse, line 272: If generalized, it could check for the source type from the config parameter.

analyses/org.eclipse.tracecompass.incubator.inandout.core/src/org/eclipse/tracecompass/incubator/internal/inandout/core/analysis/InAndOutDataProviderFactory.java
bhufmann: I created https://github.com/eclipse-tracecompass/org.eclipse.tracecompass/pull/257 and apply in this PR.

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/services/DataProviderService.java
```diff
@@ -1394,7 +1394,16 @@ public Response deleteDerivedProvider(
...
             }
+
+            // Get all DP descriptors before they are removed
+            List<IDataProviderDescriptor> allDerivedDps = manager.getAvailableProviders(experiment, derivedDescriptor.getConfiguration());
+
+            // Clean-up configuration
             configurator.removeDataProviderDescriptor(experiment, derivedDescriptor);
+
+            // Remove any derived data provider instances stored in the manager
+            allDerivedDps.forEach(desc -> manager.removeDataProvider(experiment, desc.getId()));
```
PatrickTasse, line 1405: I found that it seems to work if you remove the data providers first and then remove the derived descriptor and configuration. Is there a reason to remove them after?

bhufmann: No there is no particular reason. I can call it first as well.

PatrickTasse (approved):

PR #181, opened 2025-04-22
server: Add logging to the trace server endpoints

kavehshahedi:
### What it does

With respect to the potential security issues, this commit tends to provide the required infrastructure for logging the incoming requests and outgoing responses that are handled by the trace server. We use the `LogUtils.traceInstant()` class to capture various information (e.g., path, IP, method, etc.) for each request/response.

#### Incoming request details:
* Request's unique ID
* HTTP method, path, query parameters
* Client IP address (including support for proxy headers)
* Headers and (truncated) body content for applicable methods

#### Outgoing response details:
* Request's unique ID (from the original request)
* HTTP status code, content type, content length
* Total request processing duration

### How to test

The trace server logging should be enabled (i.e., by running it with proper logging arguments). From that point, if the permission for the `Level.INFO` log level is enabled, the server will capture each request/response information in the desired logging location (i.e., from the logging properties file).

### Follow-ups

Extra information or specific endpoint filtering might be applied based on the requirements.

### Review checklist

### DIFF
--- a/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/webapp/RequestResponseLogger.java
+++ b/trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/webapp/RequestResponseLogger.java
@@ -1,0 +1,285 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Ericsson
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License 2.0 which
+ * accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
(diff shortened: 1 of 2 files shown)


MatthewKhouzam (approved):
This looks good to me!

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/webapp/RequestResponseLogger.java
```diff
@@ -0,0 +1,303 @@
...
+
+        // Add body content (if any)
+        if (!bodyContent.isEmpty()) {
+            String truncatedBody = bodyContent.length() > MAX_BODY_LOG_LENGTH
+                    ? bodyContent.substring(0, MAX_BODY_LOG_LENGTH) + "..." //$NON-NLS-1$
+                    : bodyContent;
+            logMessage.append(" | Body: ").append(truncatedBody); //$NON-NLS-1$
+        }
+
+        LOGGER.log(Level.INFO, logMessage.toString());
```
MatthewKhouzam, line 140: LogUtils.traceInstant()

MatthewKhouzam (approved):

PatrickTasse (changes requested):

PR #182, opened 2025-04-29
inandout: Fix errors in JSON schema

PatrickTasse:
### What it does

Remove required name in schema. The name is required by TSP protocol but is not part of the schema's
properties object.

The 'required' attribute should be outside the properties object.

### How to test

Validate a JSON configuration without name field inside properties/parameters object.

### Follow-ups

N/A

### Review checklist


bhufmann (approved):
Makes sense. Thanks for fixing this.

PR #192, opened 2025-05-29
Fix tsp bugs

bhufmann:
### What it does

- swagger: Allow double and remove boolean in OutputElementStyle
Previously only float values were allowed. Don't restrict to float and allow Double as well. Note when deserializing JSON strings a floating point value will be deserialized to Double, e.g. when using Jackson deserializer (Java) or in Javascript. Remove boolean because it can be expressed as String and boolean have no equivalent type in CSS where the StyleProperties are inspired of.

- server: Only serialize supported value types in OutputElementStyle
This will make sure that only objects that comply with the TSP specification are serialized. Added unit tests to verify this.

### How to test
Successful execution of newly added test case

### Follow-ups

TSP spec. update

### Review checklist


trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core.tests/src/org/eclipse/tracecompass/incubator/trace/server/jersey/rest/core/tests/webapp/OutputElementStyleSerializerTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+        OutputElementStyle testStyle = new OutputElementStyle(INVALID_STYLE_NAME, invalidStyles);
+        String json = fMapper.writeValueAsString(testStyle);
+
+        OutputElementStyleStub deserialized = fMapper.readValue(json, OutputElementStyleStub.class);
+        assertEquals(testStyle.getParentKey(), deserialized.getParentKey());
+        assertTrue(deserialized.getStyleValues().isEmpty());
+    }
+
+    private static void veryStyles(OutputElementStyle testStyle, OutputElementStyleStub deserialized) {
```
PatrickTasse, line 125: verifyStyles?

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/webapp/OutputElementStyleSerializer.java
```diff
@@ -38,11 +40,30 @@ public OutputElementStyleSerializer() {
...
+    @SuppressWarnings({ "nls", "null" })
     @Override
     public void serialize(@NonNull OutputElementStyle value, JsonGenerator gen, SerializerProvider serializers) throws IOException {
         gen.writeStartObject();
-        gen.writeStringField("parentKey", value.getParentKey()); //$NON-NLS-1$
-        gen.writeObjectField("values", value.getStyleValues()); //$NON-NLS-1$
+        gen.writeStringField("parentKey", value.getParentKey());
+
+        // Verify the type of style values. Make sure only supported types are returned.
+        Map<String, Object> newValues = new HashMap<>();
```
PatrickTasse, line 54: Maybe could use startObject/writeField/endObject pattern to write directly without needing copied map, and preserve original map order.

PatrickTasse (changes requested):

trace-server/org.eclipse.tracecompass.incubator.trace.server.jersey.rest.core/src/org/eclipse/tracecompass/incubator/internal/trace/server/jersey/rest/core/webapp/OutputElementStyleSerializer.java
bhufmann: Ok. This will avoid looping through the map multiple times. The order doesn't really matter, unless the input map is a LinkedHashMap that preserves the insertion order.

PatrickTasse (approved):

PR #205, opened 2025-07-09
swagger: replace deprecated required field with requiredMode

bhufmann:
### What it does

swagger: replace deprecated required field with requiredMode. This removes all related deprecated warnings.

### How to test

Generate yaml file as described in step 4 of: https://github.com/eclipse-cdt-cloud/trace-server-protocol?tab=readme-ov-file#generate-the-api. The specification should have the same content in respect to required fields.

### Review checklist


MatthewKhouzam (approved):
I approve this change, i will now check if any required are lingering. i.e. this change is correct, but it may not be complete. Let's merge it.

PR #258, opened 2025-11-14
traceevent: Handle pytorch traces

arfio:
### What it does

Make use of the event path, changes the logic to handle array and object trace. Add a test using a small sample of stable diffusion trace.

### How to test

Open any pytorch trace.

### Follow-ups

The previous patch was also modifying how the trace events handle tids, but I did not have issues running the example stable diffusion trace.

### Review checklist

- [ ] As an author, I have thoroughly tested my changes and carefully followed the instructions in this template


TheMatthew:
image
In case you were wondering what the results look like.

Issue #239, opened 2025-09-30
ftrace: Suspected parser errors cause majority of events after a certain point to not be listed


felix-khlmnn:
Hello everybody,

### Bug Description:

Some time ago I have documented a bug where the parsing of ascii ftrace files stop seemingly at random when going through the file. The result is that only the events up until this stop get displayed.

I have accidentally filed the same bug report for the main repo because I thought the ftrace component is included there and not here.

Further details can be found in #eclipse-tracecompass/org.eclipse.tracecompass#193.

The issue appears to be due to an event name starting with opening parentheses "(" which in turn seems to terminate the parsing process.

I'd like to work on a fix concerning this issue. Before I start my work on it, I'd like to know whether this is a known issue that you want to have fixed :)

### Steps to Reproduce

1. Download the ZIP file below
2. Transform the binary file "long_program.dat" into a text file with `trace-cmd report > long_program.txt`
3. Load it into a Trace Compass project and compare the number of events of the binary file and the ascii file (Should be different due to error described above)
4. In the ascii file, search and replace all occurrences of "(udev-worker)" with "udev-worker"
5. Reload ascii file, compare again: Event number should be equal

[Binary Trace File](https://github.com/user-attachments/files/18110765/tracecompass-issue-193.zip)

Thanks in advance

Felix

felix-khlmnn:
On further analysis, I'm sure that it has something to do with the Regex here, which decided whether a line is part of a valid ftrace or not: [Regex line](https://github.com/eclipse-tracecompass-incubator/org.eclipse.tracecompass.incubator/blob/885c8015d130aee3256c08127d6dcb17cec5b0fc/tracetypes/org.eclipse.tracecompass.incubator.ftrace.core/src/org/eclipse/tracecompass/incubator/internal/ftrace/core/event/IGenericFtraceConstants.java#L60)

felix-khlmnn:
I have tested with regex101 and the issue seems to be in the way the "data" capture group is defined. 

When we switch `(?<data>[^\\)]*)(\\))?` with `(?<data>.*)$` regex101 gets a full match with the data in the correct fields. 
I'll try to implement this soon and test it.

nhths:
Any updates here?

felix-khlmnn:
Hey, I'm still pursuing this issue, hadn't had the time to deal with it yet due to projects in my day job.

[...]

PR #262, opened 2026-01-05
ftrace: Update ftrace event regex

felix-khlmnn:
### What it does

Fixes: #239 

As described in issue #239, the regex used for parsing ftrace events currently fails to correctly parse text files under specific circumstances. I have found that to be the case with event names that start with opening parentheses. It only occurs when parsing text files, which was necessary in my case.

In my case, the event `(udev-worker)` caused the parsing of the file to be halted. It only managed to fully parse the file when I changed every occurence of `(udev-worker)` to `udev-worker`. This led me to assume that the parentheses mess with the parsing of the data fields.

Using regex101, I have adopted the regex to correctly parse the trace provided in the "How to test" section. No further testing could be done yet as explained in the issue.

### How to test

The testing can be done using an example of a trace I have provided [here](https://github.com/user-attachments/files/18110765/tracecompass-issue-193.zip).

1. Download the ZIP file from the link above.
2. Transform the binary file "long_program.dat" into a text file with `trace-cmd report -i long_program.dat > long_program.txt`
3. Load both the binary (dat) and the ascii (txt) file into Trace Compass
4. Compare the number of events, if equal, parsing worked correctly.

### Follow-ups

If I encounter similar issues I will report them in reference to this topic and update the regex if appropriate.

### Review checklist

- [X] As an author, I have thoroughly tested my changes and carefully followed the instructions in this template
(To the best of my ability, see linked issue)

## Summary by CodeRabbit

* **Bug Fixes**
  * FTRACE event parsing now captures the full remainder of each event line, ensuring trailing content (including previously excluded characters) is consistently extracted and displayed in trace views.

* **Tests**
  * Added a parsing test covering lines that start with parentheses to validate correct field extraction and guard against regressions in this edge case.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/tracetypes/org.eclipse.tracecompass.incubator.ftrace.core.tests/src/org/eclipse/tracecompass/incubator/ftrace/core/tests/event/FtraceFieldTest.java
+++ b/tracetypes/org.eclipse.tracecompass.incubator.ftrace.core.tests/src/org/eclipse/tracecompass/incubator/ftrace/core/tests/event/FtraceFieldTest.java
@@ -192,6 +192,26 @@ public class FtraceFieldTest {
     }
 
     /**
+     * Testing of parse line with a line that starts with parentheses
+     */
+    @Test
+    public void testParseStartsWithParentheses() {
+        String line = "(udev-worker)-299   [000] d.h1.    73.294605: sched_stat_runtime:   comm=systemd-udevd pid=299 runtime=65790 [ns] vruntime=6269486605 [ns]";
+
+        GenericFtraceField field = GenericFtraceField.parseLine(line);
+
+        assertNotNull(field);
+        assertEquals(4, field.getContent().getFields().size());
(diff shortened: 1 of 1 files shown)


felix-khlmnn:
> Let's wait for CI thought.

Had a minor mistake in the test implementation, the test itself should be correctly implemented now.

MatthewKhouzam:
I checked the code...
`        String line = "test/1-1316  [005] .......   713.920983: sys_recvmsg(fd: 3, msg: 7ffe3bd38070, flags: 0)";`
Parses flags as the string '0)' keeping the closing bracket.

This is because of the greedy .*. I am tempted to implement a basic bracket parser. i.e. a +1 on ( and -1 on ) to reliably fix it. But it may be overkill.

